PR #6, opened 2017-11-15
Feat/edit profile

Jephuff:


package.json
```diff
@@ -5,13 +5,15 @@
...
+    "formik": "false0.10.4",
     "freactal": "^1.1.1",
     "glamor": "^2.20.40",
     "jwt-decode": "^2.2.0",
     "react": "^16.0.0",
     "react-dom": "^16.0.0",
     "react-scripts-ts": "2.8.0",
-    "recompose": "^0.26.0"
+    "recompose": "^0.26.0",
+    "url-join": "false2.0.2"
```
cheapsteak: Check this

src/components/Login.tsx
```diff
@@ -88,7 +89,11 @@ class Component extends React.Component {
     if (response.status === 200) {
       const jwt = response.data;
       const user = jwtDecode(jwt).context.user;
-      await props.effects.setUser(user);
+      // TODO: use id instead of email?
+      const egoId = user.email;
```
cheapsteak: Yes, if we can use IDs now then let's do that

tslint.json
```diff
@@ -1,99 +1,76 @@
 {
-    "extends": ["tslint-react"],
```
cheapsteak: Were there actual changes or are these all prettier diffs?

package.json
Jephuff: oops, fixed the config

tslint.json
Jephuff: there were actual changes. I can separate this into a another pr though.

cheapsteak: Nah, doesn't matter at this stage, just curious what they were
Feel free to merge

src/components/Login.tsx
Jephuff: Ids aren't returned in the jwt, so it would have to be a separate call right now. 
Could we add the id to the jwt @ra-ver ?

tslint.json
Jephuff: 👍 line length and semi were two of them.

src/components/Login.tsx
cheapsteak: It's being returned at the root level under as "sub"

Jephuff: oh, ok! so sub === id?

cheapsteak: Yup

PR #14, opened 2017-11-20
allow overriding the api

Jephuff:


cheapsteak:
could you do the same for REACT_APP_PROFILE_API?

Jephuff:
yep! was just doing that :)

PR #35, opened 2018-02-01
add arranger

Jephuff:


src/components/FileRepo.js
```diff
@@ -17,6 +25,32 @@ const FileRepo = ({ state, effects }) => {
         )}
       </pre>
       {!state.loggedInUser && <div>Seeing files as annon</div>}
+      <Arranger
+        index="file"
```
andricDu: Will this eventually be read in from env?

Jephuff: It can be, but the site will have multiple indexes, so if it is an environment variables it would probably be something like `REACT_APP_FILE_INDEX`.

PR #47, opened 2018-02-15
✨Feat/advanced facet view

hlminh2000:


src/components/FileRepo.js
```diff
@@ -37,7 +39,7 @@ const arrangerStyles = css`
   }
 `;
 
-const AggregationsWrapper = props => {
+const AggregationsWrapper = ({ showAdvancedFacetView = () => {}, ...props }) => {
```
Jephuff: You can wrap AggregationsWrapper with `injectState` so you don't need to pass a different callback.

src/components/FileRepo.js
```diff
@@ -105,7 +107,7 @@ const customTableTypes = {
   },
 };
 
-const FileRepo = ({ state, effects, ...props }) => {
+const FileRepo = ({ state, effects, showAdvancedFacetView = () => {}, ...props }) => {
```
Jephuff: this `showAdvancedFacetView` isn't used is it?

src/components/FileRepo.js
```diff
@@ -145,11 +146,26 @@ const FileRepo = ({ state, effects, ...props }) => {
                 streamData={props.streamData(props.index, props.projectId)}
               />
             </div>
+            {state.modalState.isShown && (
+              <AdvancedFacetViewModal
+                sqon={selectionSQON}
```
Jephuff: This should propbably be `props.sqon`. `selectionSQON` is the current rows of the table that are selected.

src/components/FileRepo.js
```diff
@@ -145,11 +146,26 @@ const FileRepo = ({ state, effects, ...props }) => {
...
+                {...{
+                  PROJECT_ID: process.env.REACT_APP_PROJECT_ID,
```
Jephuff: Just spread props and you will get projectId, index and sqon. You will need to update advanced facet to accept camelCase.
PROJECT_ID === props.projectId
ES_INDEX === props.index

API_HOST:  Inside advanced facet use utils/api. It will handle the api for you.
ES_HOST: You don't need ES_HOST in the graphql calls anymore, so you can remove this one.

src/components/FileRepo.js
```diff
@@ -145,11 +146,26 @@ const FileRepo = ({ state, effects, ...props }) => {
...
+            )}
           </div>
         );
       }}
     />
   );
 };
 
-export default enhance(FileRepo);
+export default provideModalState(enhance(FileRepo));
```
Jephuff: Put this in the compose call above
```
const enhance = compose(provideModalState, injectState);
```

src/components/AdvancedFacetViewModal/index.js
```diff
@@ -0,0 +1,109 @@
+import * as React from 'react';
+import { injectState } from 'freactal';
+import { compose } from 'recompose';
+
+import { AdvancedFacetView } from '@arranger/components/dist/Arranger';
+import { provideLocalSqon } from 'stateProviders';
+const LiveAdvancedFacetView = AdvancedFacetView;
+
+const STYLE = {
+  ADVANCED_FACET_OVERLAY: {
```
Jephuff: stick with camelCase for these (the variable top object can be `styles`)

src/components/AdvancedFacetViewModal/index.js
```diff
@@ -0,0 +1,109 @@
...
+  onSqonSubmit = () => {},
+  ...props
+}) => (
+  <div style={STYLE.ADVANCED_FACET_OVERLAY} onClick={() => closeModal()}>
+    <div style={STYLE.ADVANCED_FACET_CONTAINER} onClick={e => e.stopPropagation()}>
+      <div style={STYLE.ADVANCED_FACET_TITLE}>All filters</div>
+      <div
+        style={{
+          ...STYLE.ADVANCED_FACET_WRAPPER,
+          flex: 1,
```
Jephuff: move these in to `ADVANCED_FACET_WRAPPER`

src/stateProviders/provideLocalSqon.js
```diff
@@ -0,0 +1,15 @@
+import { provideState } from 'freactal';
+
+export default provideState({
+  initialState: props => ({
+    advancedFacetSqon: null,
```
Jephuff: These shouldn't be specific to the advanced facets, make it something like localSQON. 
It would probably also initialize to props.sqon.

src/components/AdvancedFacetViewModal/index.js
```diff
@@ -0,0 +1,109 @@
...
+          >
+            View Results
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+);
+
+export default provideLocalSqon(enhance(AdvancedFacetViewModal));
```
Jephuff: Move `provideLocalSqon` to the main compose call (before inject)

src/components/AdvancedFacetViewModal/index.js
```diff
@@ -0,0 +1,109 @@
...
+          ...STYLE.ADVANCED_FACET_WRAPPER,
+          flex: 1,
+          display: 'flex',
+        }}
+      >
+        <div style={{ position: 'relative', flex: 1 }}>
+          <LiveAdvancedFacetView
+            {...{
+              ...props,
+              sqon: advancedFacetSqon || props.sqon,
```
Jephuff: `null` would be a valid state for local sqon. If you initialize localsqon with the sqon passed in (inside the provider), you can remove this conditional.

src/components/FileRepo.js
```diff
@@ -145,6 +146,20 @@ const FileRepo = ({ state, effects, ...props }) => {
                 streamData={props.streamData(props.index, props.projectId)}
               />
             </div>
+            {state.modalState.isShown && (
+              <AdvancedFacetViewModal
+                {...{
+                  ...props,
+                  PROJECT_ID: props.projectId,
+                  ES_INDEX: props.index,
+                  API_HOST: process.env.REACT_APP_API,
```
hlminh2000: netlify seems to be missing this env variable. Is this something we can configure @Jephuff ? :)

Jephuff: It was renamed to `REACT_APP_ARRANGER_API`.

You shouldn't need to pass it in though, if AdvancedFacetViewModal uses utils/api to fetch data.

hlminh2000: @Jephuff :) Ok I refactored I removed the unneeded stuff and refactored Arranger side as part of this pull request: https://github.com/overture-stack/arranger/pull/165

PR #48, opened 2018-02-15
💄 style tweaks

Jephuff:


alex-wilmer:
I'm hoping we can move towards making the "page" not scroll at all, and just have the sidebars and table scroll independently, but out of scope of this small change

PR #70, opened 2018-02-18
✨ make admin dash part of app

alex-wilmer:
need to fix a few things, but adding the admin dash to the actual portal instance is a *much* better idea than having it be a separate application. This way we can define kf specific props and even share state between the dash and the rest of the portal like "the current version" instead of requiring it from env.

also this fixes the broken storybook routing


alex-wilmer:
image

PR #75, opened 2018-02-21
✨ use download endpoint and manifest participant only

Jephuff:
~depends on https://github.com/overture-stack/arranger/pull/176~
depends on https://github.com/overture-stack/arranger/pull/214


src/components/Stats/index.js
```diff
@@ -0,0 +1,206 @@
+import React from 'react';
+import { withProps } from 'recompose';
+import filesize from 'filesize';
+import { get } from 'lodash';
+import Stats from './Stats';
+
+export const FileRepoStats = withProps(() => ({
```
alex-wilmer: This component.. its both the stats query and the presentation right? can we decouple that? we need to get those agg stats into dynamo db as well

Jephuff: Sure, what are you thinking? `Stats` is the presentation + data fetching. But the query definitely is just a config thats passed to it. So we could easily move those to a separate file like `predefinedQueries` or something.

PR #76, opened 2018-02-21
:floppy_disk: persist jwt to localstorage

cy:
requested by @joneubank. First login will save JWT to localstorage, next loads will attempt to restore user from the saved JWT.

r? @Jephuff, don't really know what I'm doing with `gapi.auth2`.

### DIFF
--- a/src/stateProviders/provideLoggedInUser.js
+++ b/src/stateProviders/provideLoggedInUser.js
@@ -1,6 +1,9 @@
 import { provideState } from 'freactal';
 import { setToken } from 'services/ajax';
 
+import { googleAppId } from 'common/injectGlobals';
+import { handleJWT } from 'components/Login';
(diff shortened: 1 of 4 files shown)


src/stateProviders/provideLoggedInUser.js
```diff
@@ -9,6 +12,24 @@ export default provideState({
     percentageFilled: 0,
   }),
   effects: {
+    initialize: effects => state => {
+      const { setToken, setUser } = effects;
+      const jwt = localStorage.getItem('EGO_JWT');
+      if (jwt) {
+        try {
+          const gapi = global.gapi;
```
Jephuff: Couldn't find anything much better. But maybe we just do this right before logout if `gapi.auth` doesn't exist?

PR #83, opened 2018-02-23
Retrieve and store integration tokens on login

joneubank:
Added Secrets service code.
On login fetch tokens for all available integrations
Store all found integration tokens in local storage
Add all tokens to state as `integrationTokens`

Constants file modified to add all integration service keys.

SETUP UPDATE: With this PR a new property is required in .env - REACT_APP_SECRETS_API . I can provide the URL for developers/running servers to use.

https://github.com/kids-first/kf-portal-ui/issues/58


joneubank:
Adopted your suggestion to make SERVICES constant and array.

PR #96, opened 2018-03-01
74 - refine login flow

cy:
- Root route changed to login page, this displays same OAuth provider login buttons as Join wiz but different text to guide UX.
- Pulled in GDC toast Notifications, renamed to Toast because there was always confusion with these and BannerNotifications in GDC.
- Toast with 'Fill out Profile or Skip and Browse Files' on complete of join wiz.
screen shot 2018-03-01 at 2 02 50 pm

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,38 +2,32 @@ import React from 'react';
 import { compose } from 'recompose';
 import { injectState } from 'freactal';
 import './App.css';
-import { provideLoggedInUser, provideModalState } from 'stateProviders';
+import { provideLoggedInUser, provideModalState, provideToast } from 'stateProviders';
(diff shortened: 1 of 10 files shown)


cy:
match toast design
screen shot 2018-03-01 at 4 29 03 pm

src/App.js
```diff
@@ -153,6 +124,7 @@ const render = ({
           >
             {modalState.component}
           </Modal>
+          <Toast {...toast}>{Object.keys(toast).length && toast.component}</Toast>
```
Jephuff: do you need to check `Object.keys(toast).length`? toast should always be an object, and Toast will handle when it's closed I think?

cy: you're right, updated

PR #98, opened 2018-03-01
♻️ Refactor/modal

hlminh2000:


cy:
could you mess with the modal footer zindex it's covering the dropdown:
screen shot 2018-03-02 at 10 13 21 am

src/components/UserProfile/index.js
```diff
@@ -69,6 +69,30 @@ export const H4 = styled('h4')`
   font-weight: normal;
 `;
 
+const submitBasicInfoForm = async (
```
cy: is this used?

src/components/Modal/index.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import Modal from 'react-modal';
+import CloseIcon from 'react-icons/lib/md/close';
+import { withTheme } from 'emotion-theming';
+import { compose, withState } from 'recompose';
```
cy: ~~withState~~

src/stateProviders/provideModalState.js
```diff
@@ -3,33 +3,27 @@ import { provideState } from 'freactal';
 export default provideState({
   initialState: props => ({
     modalState: {
+      title: null,
       isShown: false,
```
cy: can delete isShown state now because https://github.com/kids-first/kf-portal-ui/pull/98/files#diff-14b1e33d5bf5649597cdc0e4f684daddR129

src/components/FileRepo.js
```diff
@@ -68,7 +68,28 @@ const AggregationsWrapper = injectState(({ state, effects, ...props }) => {
...
+          onClick={() =>
+            effects.setModal({
+              title: 'All Filters',
+              component: (
+                <AdvancedFacetViewModalContent
+                  {...{
+                    ...props,
+                    closeModal: effects.unsetModal,
+                    onSqonSubmit: ({ sqon }) => {
+                      console.log('sqon: ', sqon);
```
cy: ~~console.log~~

src/components/AdvancedFacetViewModal/style.css
```diff
@@ -1,32 +1,10 @@
...
-  height: 800px;
-  position: relative;
-  background: white;
-  border-radius: 5px;
-  display: flex;
-  flex-direction: column;
-  overflow: hidden;
-}
-.advacnedFacetsContainer .advancedFacetsWrapper {
+.advancedFacetsWrapper {
```
cy: can we try to use `emotion` css instead of a mix of real css and `emotion`? 😅Unless these classes are set in arranger?

src/components/AdvancedFacetViewModal/style.css
```diff
@@ -1,32 +1,10 @@
...
   flex: 1;
   display: flex;
-  padding: 20px;
-  padding-bottom: 0px;
+  min-height: 700px;
+  max-height: 800px;
 }
-.advacnedFacetsContainer .advancedFacetsTitle {
+.advancedFacetsTitle {
```
cy: wait these classes aren't still used because you deleted it here, https://github.com/kids-first/kf-portal-ui/pull/98/files#diff-81ef1d93b3acd5aa96fc27e3be1bd3a3L34 right? And now using https://github.com/kids-first/kf-portal-ui/pull/98/files#diff-b17c47ef8137161ae8a178243850f1c4R12 for the header style?

src/App.js
```diff
@@ -127,32 +127,14 @@ const render = ({
...
-                width: '55%',
-                boxShadow: 'rgba(0, 0, 0, 0.5) 0px 5px 15px',
-                overflow: 'visible',
-              },
+            {...{
+              unsetModal: effects.unsetModal,
+              setModal: effects.setModal,
+              title: modalState.title,
+              content: modalState.component,
+              appElement,
```
Jephuff: Is app element used? 
having the default value be a document selector is going to be slow, so if it's not used lets remove it. if it is, maybe we can refactor to not need it :D

src/App.js
```diff
@@ -127,32 +127,14 @@ const render = ({
...
-                border: '1px solid rgb(204, 204, 204)',
-                background: 'rgb(255, 255, 255)',
-                borderRadius: '4px',
-                margin: '30px auto',
-                width: '55%',
-                boxShadow: 'rgba(0, 0, 0, 0.5) 0px 5px 15px',
-                overflow: 'visible',
-              },
+            {...{
+              unsetModal: effects.unsetModal,
```
Jephuff: Probably makes sense to make modal get all these arguments from the provider itself. so this could end up being just `<Modal />`

src/App.js
hlminh2000: @Jephuff it's fed into react-modal I believe. I can have it wrapped in a service function and load it in the Modal component directly

Jephuff: ahh I see. So doing that is still going to run that selector every time it renders. 
If you can get away with passing it the `.app` div instead of the `#root` div you can use a ref to get the element and pass it in.

Otherwise you can just run that selector once and store the dom node, it won't change (it could maybe go in `injectGlobals`

src/App.js
hlminh2000: oh but other components need to call the `setModal` and `unsetModal`, so don't we need to provide the state from the top level?

Jephuff: oops, I should have been clearer. 
You would leave the provider where it is. But just add `injectState` to the modal. Then it will get all those values as well.

hlminh2000: oh yeah ok now I see how this works :)))

PR #108, opened 2018-03-06
⬆️ Arranger 0.1.12

hlminh2000:


Jephuff:
Can you graphqlField to the arranger components as well? there may need to be a couple more arranger updates to accept it

PR #113, opened 2018-03-07
Cavatica Widget in File Repo Sidebar

joneubank:
Providing Cavatica interaction in File Repo sidebar.

 * Connect to Cavatica button if not connected
 * Project list if Connected
 * Create a project
 * Copy files to project


src/common/injectGlobals.js
```diff
@@ -19,9 +19,12 @@ if (personaApiOverride) {
   global.log('warning: using persona api override');
 }
 
+export const arrangerProjectId = process.env.REACT_APP_PROJECT_ID;
```
Jephuff: wanna replace `process.env.REACT_APP_PROJECT_ID` with this in FileRepo.js?

src/components/FileRepoSidebar.js
```diff
@@ -178,27 +178,6 @@ export default ({ projectId, style, sqon, ...props }) => (
...
-      <img
-        alt=""
-        src={cavaticaLogo}
-        css={`
-          width: 28px;
-          margin-right: 7px;
-        `}
-      />Export files to Cavatica
-    </button>
+    <CavaticaSidebar {...props} />
```
Jephuff: Could you rename this to something like `CavaticaExportButton`. the external link could be moved to be a sibling here.

src/services/cavatica.js
```diff
@@ -80,16 +82,70 @@ export const getProjects = async () => {
...
 };
+
+/**
+ * Requires Cavatica File IDs as returned from convertGen3FileIds
+ *
+ * project is the project id, not name (ie. userId/projectName)
+ * ids is an array of strings of the ids to copy
+ */
+export const copyFiles = async ({ project, ids }) => {
+  const { data: { items } } = await ajax.post(cavaticaApiRoot, {
```
alex-wilmer: if `ajax.post` is unsuccesful, will this crash trying to access `data.items` ?

src/components/cavatica/CavaticaAddProject.js
```diff
@@ -0,0 +1,100 @@
...
+            `}
+          >
+            Create a new Cavatica Project
+          </div>
+          <div css={``}>Enter a name for your project:</div>
+          <input
+            className="textInput"
+            type="text"
+            placeholder="Project Name"
+            onChange={e => setProjectName(e.target.value)}
```
Jephuff: Might be a good idea to add `value={projectName}` incase we want to populate that down the line.

src/components/cavatica/CavaticaAddProject.js
```diff
@@ -0,0 +1,100 @@
...
+          />
+        </div>
+        <div className="footer">
```
Jephuff: Not sure if you saw it, but in Modal/index.js there is a `ModalFooter` component. I think you should be able to use that here

src/components/cavatica/CavaticaSidebar.js
```diff
@@ -0,0 +1,138 @@
...
+    :disabled {
+      cursor: default;
+      color: ${theme.greyScale2}
+      background-color: ${theme.greyScale5}
+    }
+  }
+`;
+
+const CavaticaSidebar = ({ state, theme, addingProject, setAddingProject, ...props }) => {
+  let content = <CavaticaConnectButton styles={styles(theme)} />;
```
Jephuff: you can pass this to className or css (`css={styles(theme)}`or `className={styles(theme)}`) and just pass className along inside the `CavaticaConnectButton` component.

src/icons/DoubleChevronLeftIcon.js
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+
+export default ({ height, width, fill = '#fff', style, className, alt = '', ...props }) => {
+  return (
+    <img
+      src={`data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 99.2 90.93"><defs><style>.cls-1{fill:${fill};}</style></defs><title>icon-double-chevron-left-grey</title><path class="cls-1" d="M38,3.05,3,38A10.55,10.55,0,0,0,3,52.88l35,35A10.42,10.42,0,0,0,52.78,73.14L25.1,45.46,52.78,17.79A10.42,10.42,0,0,0,38,3.05"/><path class="cls-1" d="M81.42,3.05l-35,35a10.55,10.55,0,0,0,0,14.83l35,35A10.42,10.42,0,0,0,96.15,73.14L68.47,45.46,96.15,17.79A10.42,10.42,0,0,0,81.42,3.05"/></svg>`}
```
Jephuff: You can import svgs . 
if you put this in a file call `DoubleChevronLeftIcon.svg` you can do 
```
import doubleChevronLeftIcon from './DoubleChevronLeftIcon.svg'
<img src={doubleChevronLeftIcon}/>
```

src/services/cavatica.js
joneubank: More than likely. Whats the recommended approach for handling failure here? 

ajax.post().then().catch() ?

alex-wilmer: you can use `try / catch` with async calls. i think that's the standard way a la https://github.com/overture-stack/arranger/blob/master/modules/server/src/index.js#L42

src/common/injectGlobals.js
joneubank: I don't follow. This is a global constant from our .env file, why would it be setup differently than the other .env variables?

src/components/FileRepoSidebar.js
joneubank: Its not just the button though, this wraps the project list and project adder... its the whole sidebar widget removed from the FileRepoSidebar component, just included inside.

src/components/cavatica/CavaticaAddProject.js
joneubank: Formatting is significantly different, I could try it didn't look promising.. functionality is very similar though, so long as its not tied into the modal state...

src/icons/DoubleChevronLeftIcon.js
joneubank: I modify the svg with ${fill} in the fill component. This is done to allow color control. Modeled after existing icons.

src/common/injectGlobals.js
Jephuff: we are using ` process.env.REACT_APP_PROJECT_ID` directly in FileRepo.js,  just now that you added it here you could update it to use this `arrangerProjectId` value instead.

src/icons/DoubleChevronLeftIcon.js
alex-wilmer: then you can mess with any attributes, normal react props style

src/components/FileRepoSidebar.js
Jephuff: `CavaticaExportWidget` maybe? 
sidebar is confusing, cause it's not a sidebar :D

src/components/cavatica/CavaticaAddProject.js
Jephuff: If it's different don't worry about it. Just wanted to mention incase it was the same and you just hadn't seen it. It's hiding :)

src/common/injectGlobals.js
joneubank: OH, understood now. Will update.

src/components/FileRepoSidebar.js
joneubank: Yes. Will rename as Widget. I like Widget.

src/icons/DoubleChevronLeftIcon.js
joneubank: thats way cooler. We have a handful of icons to remake this way then.

src/components/FileRepoSidebar.js
joneubank: Renamed as suggested.

src/components/cavatica/CavaticaSidebar.js
joneubank: Tried this but couldn't get the styles to apply correctly. Would like to review this setup with you next week.

joneubank: Just keeping it this way for now since it is working and about the same code wise.

joneubank:
Leaving the icons as written since its consistent with existing svg implementations. We can edit the whole set as a separate ticket.

Addressed all other comments.

joneubank:
Ready for another review and merge. Thanks!

PR #129, opened 2018-03-08
update stats queries

Jephuff:
not sure how many participants there should be, but now it's 100
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "@arranger/components": {
-      "version": "0.1.21",
-      "resolved": "https://registry.npmjs.org/@arranger/components/-/components-0.1.21.tgz",
-      "integrity": "sha512-zxoGeTsLjQrz/h0rawJRuIKkuOfPphwH02MVDbn0afpo9spTvlLe1M+uAF3mH1eTP7jmHXI55Es3SVZSjHuN1g==",
+      "version": "0.1.25",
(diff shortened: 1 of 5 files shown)


andricDu:
Still not correct.

andricDu:
@Jephuff the aggs being generated by the middleware seem to default to a size:100

Jephuff:
that explains the 100 :) 
that will be a fix in arranger. I'll take a look.

andricDu:
I'm guessing the number in the UI is just a count of the returned buckets which totally makes sense as its the sum of unique terms.

Jephuff:
I think this is what we really want https://www.elastic.co/guide/en/elasticsearch/guide/current/cardinality.html
I don't know if there is a good way of getting all buckets. I'll update with a bad way for now and look more into that.

andricDu:
Issue with cardinality is that it is not an exact count. We went down that road in ICGC and decided it wasn't good as the numbers in the UI were not accurate. I'll double check to see what we did there just so we can have another reference.

andricDu:
Yea it looks like even in ICGC to produce the summary data we counted the number of buckets https://github.com/icgc-dcc/dcc-portal/blob/develop/dcc-portal-server/src/main/java/org/icgc/dcc/portal/server/repository/FileRepository.java#L753

And we currently use a large constant for that https://github.com/icgc-dcc/dcc-portal/blob/develop/dcc-portal-server/src/main/java/org/icgc/dcc/portal/server/repository/FileRepository.java#L161

andricDu:
@Jephuff u da mvp
image

PR #155, opened 2018-03-09
Feature/gen3integration

ra-ver:
End to end file downloads from Gen3


src/components/FileRepoSidebar.js
```diff
@@ -30,6 +37,17 @@ const styles = {
...
 };
 
+const getGen3UUIDs = async arrangerIds => {
+  const fileData = await getFilesById({ ids: arrangerIds, fields: ['uuid'] });
+  return fileData.map(file => file.node.uuid);
+};
+
+const downloadFile = async (arrangerIds, gen3Key) => {
+  let files = await getGen3UUIDs(arrangerIds);
+  console.log(files);
```
alex-wilmer: remove please

src/components/FileRepoSidebar.js
```diff
@@ -30,6 +37,17 @@ const styles = {
...
+  let fileUUID = files && files.length > 0 ? files[0] : '';
+  return downloadFileFromGen3(gen3Key, fileUUID);
```
alex-wilmer: does it make sense to still call this function with an empty string for `fileUUID`?

src/components/FileRepoSidebar.js
```diff
@@ -19,6 +23,9 @@ import {
 } from '../services/downloadData';
 import FamilyManifestModal from './FamilyManifestModal';
 
+let gen3Key = '';
```
alex-wilmer: lets remove these module globals and directly use the eventual values

src/components/FileRepoSidebar.js
```diff
@@ -67,78 +85,36 @@ const Heading = styled('div')`
...
-  <div
-    css={`
-      ${styles.container} ${style};
-    `}
-  >
-    <Heading>
-      File Action <InfoIcon />
-    </Heading>
+const FileRepoSidebar = ({ state, projectId, index, style, sqon, effects, theme, ...props }) => {
+  gen3Key = state.integrationTokens[GEN3];
```
alex-wilmer: re: above comment.. don't bother with this step. use `state.integrationTokens[GEN3];` below

alex-wilmer: or at the very least, initialize `gen3Key` here to avoid unnecessary mutaton

src/components/FileRepoSidebar.js
```diff
@@ -147,46 +123,184 @@ const FileRepoSidebar = ({ projectId, index, style, sqon, effects, ...props }) =
...
+                  Selected File
+                </Heading>
+                <div
+                  css={`
+                    display: flex;
+                    margin-bottom: 13px;
+                  `}
+                >
+                  <LoadingOnClick
+                    onClick={async () => {
```
alex-wilmer: `async` but no `await`. this can be very confusing for readers. you should either pick using `async`  with `await` and don't call `.then` or don't bother with `async` at all. I know that sometimes you'll see our code mix them, but it's in places where it's much easier to reason about why it's done like that.

example:

```
let response = await fetch(url).then(r => r.json())
```
it's mixed there, but it's because it's easy to follow.
the same can, and often is written as

```
let response = await fetch(url)
let data = await response.json()
```

which I think is totally fine. my suggestion for you is to choose one style to get more familiarized with it

src/components/FileRepoSidebar.js
```diff
@@ -147,46 +123,184 @@ const FileRepoSidebar = ({ projectId, index, style, sqon, effects, ...props }) =
...
+                                  <div>Unable to download file</div>
+                                  <div
+                                    css={`
+                                      color: 'red';
+                                      margin-bottom: 20px;
+                                      padding: 20px;
+                                    `}
+                                  >
+                                    <span>
+                                      Your Account doesn't have required permission to download this
```
andricDu: Your Account ~doesn't~ `does not` have `the` required permission to download this

src/components/FileRepoSidebar.js
ra-ver: how much ever i tried avoiding global for these, i believe these variable needs to be in global scope because of the way ColumnState component handles state. Moved the declaration though to avoid mutation.

src/components/FileRepoSidebar.js
```diff
@@ -140,46 +114,183 @@ const FileRepoSidebar = ({ projectId, index, style, sqon, effects, ...props }) =
...
+                                  <div>Unable to download file</div>
+                                  <div
+                                    css={`
+                                      color: 'red';
+                                      margin-bottom: 20px;
+                                      padding: 20px;
+                                    `}
+                                  >
+                                    <span>
+                                      Your Account does not have required permission to download
```
andricDu: Still missing an article. `required` makes  the noun `permission` specific and requires a definite article. 

Change to:
Your Account does not have **the** required permission to download this file.

alex-wilmer: can we lowercase "Account" as well? capital seems out of place there

ra-ver: Sorry about that. Updated the message

PR #156, opened 2018-03-09
✨ hack for cross index until proper solution

Jephuff:


andricDu:
What issue does this relate to?

alex-wilmer:
@andricDu in the file repository, when you select aggs, the field names are file index fields. the generated  sqon is passed along to the participant download button so in order for the same sqon to be valid on the participant index we need to drop `participant.` from the beginning of sqon fields

PR #157, opened 2018-03-09
share / save query

alex-wilmer:
- Some refactoring here to mention.. the App level context providers have been moved out into their own component since React 16 portals don't get rendered in the main tree and might need access to context like router or emotion theme.

- stats config objects lifted to functions to handle field renaming

PR adds share and save dialogs in the file repo page

also this adds the env REACT_APP_SHORTURL_API=https://13gqusdt40.execute-api.us-east-1.amazonaws.com/Dev 

(notice the capital D in `/Dev`)


alex-wilmer:
https://github.com/kids-first/kf-internal-tracker/issues/61

PR #158, opened 2018-03-10
♻️ cleanup App.js

alex-wilmer:
Moves some components in around to make App.js a little cleaner


andricDu:
@alex-wilmer I saw on netlify that the file search wasn't loading, and I was able to reproduce it locally after this refactor.

PR #166, opened 2018-03-11
white button background for dropdowns

andricDu:
Before: 
image

After:
image

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -11,3 +11,11 @@
 #root {
   height: 100%;
 }
+
+.dropDownHeader {
+  z-index: 100;
+}
+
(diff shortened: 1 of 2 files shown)


alex-wilmer:
why this file? src/components/AdvancedFacetViewModal/style.css

is this part of the advanced facet view or something more global like?

alex-wilmer:
for completely global stuff ideally it would be in theme/defaultTheme and used where needed. for a quick global fix app.css in the root is probably best

andricDu:
Wanted it next to `dropDownHeader`. As to where both of those belong... \o/

PR #187, opened 2018-03-12
removes updates of filestats for row selection

andricDu:


Jephuff:
Is this because the stats go to 0?
this should fix that https://github.com/kids-first/kf-portal-ui/pull/188

andricDu:
@Jephuff was a hero and fixed the bug. Killing this PR.

PR #210, opened 2018-03-12
Dashboard profile colours switch on role

lepsalex:


Jephuff:
@lepsalex do you have prettier installed? Looks like maybe you are using vscode without the prettier extension.

PR #218, opened 2018-03-13
✨ use existing column configs

Jephuff:


hlminh2000:
requires https://github.com/overture-stack/arranger/pull/319

PR #221, opened 2018-03-13
✨ remove some socket props

Jephuff:
for if https://github.com/overture-stack/arranger/pull/267 is merged

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@arranger/components": "^0.3.1",
+    "@arranger/components": "^0.3.3",
```
hlminh2000: shouldn't this be `0.3.4` now?

PR #253, opened 2018-03-27
✨ update completion circle

Jephuff:


src/components/CompleteOMeter.js
```diff
@@ -2,76 +2,37 @@ import * as React from 'react';
...
-import { get } from 'lodash';
 
-import { ROLES } from 'common/constants';
-
-const CompleteOMeter = compose(withTheme)(({ theme, role, percentage, className }) => (
+const CompleteOMeter = compose(withTheme)(({ theme, percentage, className }) => (
   <div
-    css={`
-      width: 310px;
+    className={css`
```
alex-wilmer: why the change to `className` ?

src/components/UserProfile/CompletionWrapper.js
```diff
@@ -0,0 +1,69 @@
...
+import { compose } from 'recompose';
+import { injectState } from 'freactal/lib/inject';
+import { VictoryPie } from 'victory-pie';
+
+const enhance = compose(withTheme, injectState);
+
+const CompletionWrapper = ({
+  completed,
+  theme,
+  state: { loggedInUser },
```
alex-wilmer: can this be abstracted a bit? maybe it doesn't need to be, but it feels weird for this particular component to care specifically about user state

alex-wilmer: looks like you're not using it actually

src/components/CompleteOMeter.js
Jephuff: ... I don't know hahaha. oops

src/components/UserProfile/CompletionWrapper.js
Jephuff: hmm yeah, it was there  when I had the profile image in there as well.  removed now!

any idea why eslint isn't warning on unused destructured arguments? Does atom warn you?

src/components/CompleteOMeter.js
Jephuff: fixed :)

PR #273, opened 2018-04-05
Copy to Cavatica in Modal with Gen3 Permissions

joneubank:
https://app.zenhub.com/workspace/o/kids-first/kf-portal-ui/issues/225
https://app.zenhub.com/workspace/o/kids-first/kf-portal-ui/issues/226

PR #276, opened 2018-04-05
Copy to Cavatica all in query if none selected

joneubank:
Can be slow if the total query size is large since we have to batch the requests to Cavatica to be 75 files at a time. Longer than that and we run into timeout issues.


joneubank:
Built on top of the previous PR, merge that first: https://github.com/kids-first/kf-portal-ui/pull/273

PR #279, opened 2018-04-06
🐛 this should rid us of the graphql 400?

jgnieuwhof:
image


jgnieuwhof:
@hlminh2000 - this _should_ be good to go now. there's a new branch of logic in there to use the entire set of family member ids (including participants) if there are no family member ids left after subtracting the participant id set, which handles that case when the particular study facet filter is applied.

hlminh2000:
Looks good except it seems if a study ID is selected, there's `dataTypes.length` remains at 0 so it keeps spinning... maybe we should have a different state?

PR #280, opened 2018-04-06
i18next translations

joneubank:
Completed all components in the /components and /components/ShareSaveQuery directories.


joneubank:
Strings that were embedded in component logic are not translated. It is possible to use the t('key') translation funciton is possible, but the scanner doesn't pick up these keys. Since the scanner overwrites our modifications it is impossible to add these translations under the current configuration.

jgnieuwhof:
will merge for now so this doesn't go stale, can be addressed after the fact

PR #281, opened 2018-04-09
Adds support for Riff with Ego

andricDu:
Dependencies for merge:
https://github.com/overture-stack/riff/issues/1
https://github.com/kids-first/kf-arranger/pull/18


jgnieuwhof:
@andricDu - looks like the diff has a bunch of formatting in it, maybe a `prettier` config issue?

andricDu:
@jgnieuwhof should be fixed now, I'll take a look at the merge conflict.

src/components/ShareSaveQuery/shortenApi.js
```diff
@@ -6,11 +6,13 @@ export default ({ stats, queryName, sqon, loggedInUser }) => {
   let { Files, Participants, Families, Size } = stats;
   let alias = queryName || sqonToName({ filters: sqon });
 
+  const jwt = localStorage.getItem('EGO_JWT');
```
jgnieuwhof: `jwt` is available in global state `loggedInUserToken`, we should probably grab it from there to maintain the source of truth.

src/stateProviders/provideSavedQueries.js
```diff
@@ -15,28 +15,39 @@ export default provideState({
     removeDeletingId: update((state, id) => ({
       deletingIds: state.deletingIds.filter(dId => dId !== id),
     })),
-    getQueries: (effects, egoId) =>
-      effects
+    getQueries: (effects, egoId) => {
+      const jwt = localStorage.getItem('EGO_JWT');
```
jgnieuwhof: same comment here (and line 35 below), should provide jwt via state provider rather than directly from localstorage

PR #286, opened 2018-04-11
✨ improved routing

hlminh2000:
two things done:

- [x] redirects bad urls to dashboard
- [x] delays rendering of login page if the user is being loaded to avoid flashing


src/stateProviders/provideLoggedInUser.js
```diff
@@ -41,7 +42,10 @@ export default provideState({
           }
         });
       }
-      return state;
+      return {
+        ...state,
+        isLoadingUser: jwt ? true : false,
```
alex-wilmer: I think `Boolean(jwt)` or `!!jwt` would be better

PR #293, opened 2018-04-17
Bug fix for svg data not being uri encoded. (non-chrome browsers)

andricDu:
Not sure if there is a better place to ensure the encoding.


alex-wilmer:
I think the best solution is to use the svg directly and not using an img tag. jonE did this here https://github.com/kids-first/kf-portal-ui/blob/master/src/icons/CircleCheckIcon.js

andricDu:
Works for me.

andricDu:
Updated to using svg elements rather than img elements per @alex-wilmer 's suggestion

cy:
so the reason I wrapped the svgs in img tags was it seemed to respect the width/height better, look at the /join page ;(
image

andricDu:
Thanks @cy I'll do that.

PR #295, opened 2018-04-17
✨ Common api fetcher

hlminh2000:
implements custom fetcher, requires arranger update (https://github.com/overture-stack/arranger/pull/332)

### DIFF
--- a/src/components/ShareSaveQuery/shortenApi.js
+++ b/src/components/ShareSaveQuery/shortenApi.js
@@ -2,31 +2,38 @@ import urlJoin from 'url-join';
 import sqonToName from 'common/sqonToName';
 import { shortUrlApi } from 'common/injectGlobals';
 
-export default ({ stats, queryName, sqon, loggedInUser }) => {
+export default ({ stats, queryName, sqon, loggedInUser, api }) => {
   let { Files, Participants, Families, Size } = stats;
   let alias = queryName || sqonToName({ filters: sqon });
 
   const jwt = localStorage.getItem('EGO_JWT');
-  // TODO: use ajax service?
-  return fetch(urlJoin(shortUrlApi, 'shorten'), {
-    method: 'POST',
-    headers: {
-      'Content-Type': 'application/json',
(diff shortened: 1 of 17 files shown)


src/components/ShareSaveQuery/shortenApi.js
```diff
@@ -2,31 +2,38 @@ import urlJoin from 'url-join';
...
-    }),
-  }).then(r => r.json());
+  });
+
+  return api
+    ? api({
+        url: urlJoin(shortUrlApi, 'shorten'),
+        body,
+      })
+    : fetch(urlJoin(shortUrlApi, 'shorten'), {
```
jgnieuwhof: I don't *think* we need this default now that api is being passed through, I'd be in favour of using `api` without the alternative

src/stateProviders/provideSavedQueries.js
```diff
@@ -15,33 +15,44 @@ export default provideState({
...
-            },
-          }),
+        .then(
+          () =>
+            api
+              ? api({
+                  url: urlJoin(shortUrlApi, 'user', egoId),
+                  method: 'GET',
+                })
+              : fetch(urlJoin(shortUrlApi, 'user', egoId), {
```
jgnieuwhof: Is this one necessary too? There's only one instance of `getQueries` being called, passing in the `api`... maintaining the pattern seems like a good idea, rather than providing defaults here & there

src/stateProviders/provideSavedQueries.js
```diff
@@ -15,33 +15,44 @@ export default provideState({
...
-            },
-          }),
+        .then(
+          () =>
+            api
+              ? api({
+                  url: urlJoin(shortUrlApi, queryId),
+                  method: 'DELETE',
+                })
+              : fetch(urlJoin(shortUrlApi, queryId), {
```
jgnieuwhof: same here?

hlminh2000:
@jgnieuwhof fallbacks are removed! :)

PR #311, opened 2018-04-23
Adds usersnap integration

bdolly:
Adds [UserSnap](https://usersnap.com/) feedback widget to all pages for anonymous and authenticated users.

### DIFF
--- a/.env.schema
+++ b/.env.schema
@@ -30,3 +30,7 @@ REACT_APP_CAVATICA_WEB=
 
 # Google maps
 REACT_APP_GOOGLE_MAPS_KEY=
+
+# Usersnap
+REACT_APP_USERSNAP_HOST=https://api.usersnap.com/load/
+REACT_APP_USERSNAP_ID=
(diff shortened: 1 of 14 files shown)


hlminh2000:
:P thanks @jgnieuwhof I was wondering bout the modal styling too

bdolly:
@jgnieuwhof working on preview. Looks good to merge to me, if you're cool with it

bdolly:
scratch that looks like something got borked on the FileRepo sidebar

jgnieuwhof:
@bdolly - interesting, wonder what's causing that. seems unrelated to this PR, we recently did merge some sidebare changes that may have introduced this. I'm ok with merging as-is and touching that up in a subsequent PR, but if you'd prefer to address now then that's cool too. there's a CSS refactor coming down the pike pretty soon, so it might also be reasonable to stash this issue & address it until after the refactor.

@hlminh2000, @bdolly  - thoughts?

hlminh2000:
yeah if I'm ok with merging this as-is myself. We might find more problems with the image capture later, in which case it's still better to have it merged in master so we can continuously test I think.

bdolly:
sounds good to me. That will give the team time to send in bugs and test the process as well

PR #314, opened 2018-04-23
uses sum of buckets doc_count

hlminh2000:
This seems to be the ultimate problem for the family members count: it should be taking sum of doc_counts, not the # of buckets

### DIFF
--- a/src/components/FamilyManifestModal/FamilyManifestModal.js
+++ b/src/components/FamilyManifestModal/FamilyManifestModal.js
@@ -276,19 +276,23 @@ const FamilyManifestModal = ({
                   ? spinner
                   : (dataTypes || []).map(bucket => {
                       const aggs = get(data, `file`);
+                      const familyMemberBuckets = get(
+                        aggs,
+                        `${bucket.key.replace(
+                          /[^\da-z]/gi,
+                          '',
+                        )}family.participants__family__family_members__kf_id.buckets`,
+                      );
+                      const familyMembersCount =
+                        familyMemberBuckets &&
+                        familyMemberBuckets.reduce((sum, bucket) => sum + bucket.doc_count, 0);
                       return (
                         <DataTypeOption
                           disabled={isDisabled}
                           key={bucket.key}
                           bucket={bucket}
                           values={values}
-                          familyMembers={get(
-                            aggs,
-                            `${bucket.key.replace(
-                              /[^\da-z]/gi,
-                              '',
-                            )}family.participants__family__family_members__kf_id.buckets.length`,
-                          )}
+                          familyMembers={familyMembersCount}
                           fileSize={get(
                             aggs,
                             `${bucket.key.replace(/[^\da-z]/gi, '')}.file_size.stats.sum`,


alex-wilmer:
so what does the `kf_id` represent here? it looks like it would be on the `family_member` which sounds like it would be unique. or is it the family `kf_id` ?

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -276,19 +276,23 @@ const FamilyManifestModal = ({
...
                   : (dataTypes || []).map(bucket => {
                       const aggs = get(data, `file`);
+                      const familyMemberBuckets = get(
+                        aggs,
+                        `${bucket.key.replace(
+                          /[^\da-z]/gi,
+                          '',
+                        )}family.participants__family__family_members__kf_id.buckets`,
+                      );
+                      const familyMembersCount =
```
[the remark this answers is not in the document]
hlminh2000: right! not sure why I didn't think of that

hlminh2000:
from my understanding, `kf_id` here is the ids of family members for a given participant. It's a "foreign key" essentially

alex-wilmer:
so if participant A has family members B and C, who are also participants, the total number of family members should be 6?

hlminh2000:
uhm that seems logically fair but I don't think we have considered that. @rosibaj is that the expected behaviour?

hlminh2000:
kinda changed the scope of the PR alittle, but now it seems to be adding up for both file counts and family members. The main changes:

- Instead of using the sqon to filter participants on from family members on ES side, it queries for participant ids, then use that to filter out participants from the family member ids,
- uses the same sqon for both file size and family members query for each data type.

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -86,68 +88,69 @@ const enhance = compose(
...
   })),
+  withProps(({ participantIdsAggregation: { data }, familyMemberIds }) => {
+    const participantIds = (get(data, 'file.aggregations.participants__kf_id.buckets') || []).map(
+      b => b.key,
+    );
+    return {
+      participantIds,
+      familyMembersWithoutParticipantIds: familyMemberIds.filter(
+        id => !participantIds.some(pId => id === pId),
+      ),
```
jgnieuwhof: this is always going to run `O(m*n)`, I _think_ lodash `difference` has an optimization for larger array sizes, should be able to just do `difference(participantIds, familyMemberIds)`

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -267,28 +274,34 @@ const FamilyManifestModal = ({
...
+                      const familyMemberBuckets = get(
+                        aggs,
+                        `${bucket.key.replace(
+                          /[^\da-z]/gi,
+                          '',
+                        )}family.participants__family__family_members__kf_id.buckets`,
+                      );
+                      const familyMembersCount = (familyMemberBuckets || []).filter(
+                        ({ key }) => !participantIds.some(id => id === key),
+                      ).length;
```
jgnieuwhof: maybe lodash difference here too:
- `difference((familyMemberBuckets || []).map(x => x.key), participantIds).length`

src/components/FamilyManifestModal/FamilyManifestModal.js
hlminh2000: haha it does read better too :D I should really start using more of lodash lodash

PR #316, opened 2018-04-24
🐛 don't set the JWT if it's invalid

jgnieuwhof:
this introduces a check to see if a user's JWT is expired before setting it in state (on load). it'll get us around the persona api errors on page load, however a subsequent PR should address persona authentication issues in the case of a token expiring within the context of a single session.

### DIFF
--- a/src/stateProviders/provideLoggedInUser.js
+++ b/src/stateProviders/provideLoggedInUser.js
@@ -3,9 +3,8 @@ import { isArray, get } from 'lodash';
 import { addHeaders } from '@arranger/components';
 import { setToken } from 'services/ajax';
 import { updateProfile, getAllFieldNamesPromise } from 'services/profiles';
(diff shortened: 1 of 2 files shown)


src/stateProviders/provideLoggedInUser.js
```diff
@@ -19,33 +18,19 @@ export default provideState({
...
             state.integrationTokens[service] = storedToken;
           }
         });
+        return { ...state, isLoadingUser: true };
       }
-      return {
-        ...state,
-        isLoadingUser: !!jwt,
-      };
+      localStorage.removeItem('EGO_JWT');
```
hlminh2000: Small thing, maybe use `effects.setToken(null)` here? It also clears out the ajax headers and all that jazz :)

jgnieuwhof: @hlminh2000 - yeah for sure, good call :)

PR #318, opened 2018-04-25
🐛 tracks down more 401 cases

hlminh2000:
Mostly those called during signups

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -24,9 +24,10 @@ import loginImage from 'assets/smiling-girl.jpg';
 import joinImage from 'assets/smiling-boy.jpg';
 import logo from 'theme/images/logo-kids-first-data-portal.svg';
 import { requireLogin } from './common/injectGlobals';
-import { initializeApi, ApiContext, withApi } from 'services/api';
+import { withApi } from 'services/api';
 
 const forceSelectRole = ({ loggedInUser, isLoadingUser, ...props }) => {
+  console.log('loggedInUser: ', loggedInUser);
   if (!loggedInUser && requireLogin) {
     return isLoadingUser ? null : (
       <SideImagePage sideImage={loginImage} {...{ ...props }} Component={LoginPage} />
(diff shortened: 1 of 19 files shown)


hlminh2000:
yeah... singleton! :D

src/App.js
```diff
@@ -139,6 +140,18 @@ const App = compose(injectState)(({ editing, setEditing, state, effects }) => {
...
+  return (
+    <ApiContext.Provider
+      value={initializeApi({
+        onUnauthorized: response => {
+          window.location.reload();
+        },
+      })}
+    >
+      <AppWithApi />
+    </ApiContext.Provider>
```
jgnieuwhof: It may make sense to put this provider nested inside the `ContextProvider` below...
```
<ContextProvider>
  <ApiContext.Provider value={...}>
    <App />
  </ApiContext.Provider>
</ContextProvider>
```

src/App.js
```diff
@@ -24,36 +24,32 @@ import loginImage from 'assets/smiling-girl.jpg';
...
   } else if (loggedInUser && (!loggedInUser.roles || !loggedInUser.roles[0])) {
     return <Redirect to="/join" />;
   } else {
-    return <Page {...{ ...props, api }} />;
+    return <Page {...{ ...props }} />;
   }
 };
 
-const App = compose(injectState)(({ editing, setEditing, state, effects }) => {
+const App = compose(injectState)(({ editing, setEditing, state }) => {
```
jgnieuwhof: I think this could be simplified:
```
const App = compose(injectState, withApi)(({ editing, setEditing, state, api }) => (
   ...
));
```

src/services/api.js
```diff
@@ -1,26 +1,29 @@
+import React from 'react';
 import { arrangerApiRoot } from 'common/injectGlobals';
 import urlJoin from 'url-join';
 import ajax from './ajax';
 
-const initializeApi = ({ onUnauthorized }) => ({
+export const initializeApi = ({ onUnauthorized }) => async ({
```
jgnieuwhof: don't think this is actually an async function (no `await` in there)

src/services/api.js
```diff
@@ -1,26 +1,29 @@
...
       }
     });
 };
-export default initializeApi;
+
+export const ApiContext = React.createContext(null);
+
+export const withApi = WrappedComponent => props => (
+  <ApiContext.Consumer>{api => <WrappedComponent {...{ api, ...props }} />}</ApiContext.Consumer>
+);
```
jgnieuwhof: jassss hoc ftw

src/stateProviders/provideLoggedInUser.js
```diff
@@ -34,7 +34,9 @@ export default provideState({
     },
     setUser: (effects, user) =>
       getAllFieldNamesPromise()
```
jgnieuwhof: this gui needs an `api` now right?

PR #319, opened 2018-04-26
🐛 the missing bucket was throwing counts off by 1

jgnieuwhof:
the missing bucket was throwing counts off by one. I've added a PR to arranger to give the gql client the ability to specify whether or not to include the missing bucket, this uses that to fix the issue.

depends on https://github.com/overture-stack/arranger/pull/346


hlminh2000:
💯

PR #324, opened 2018-04-30
🐛 response isn't in an obj

jgnieuwhof:
was getting a 'cannot read property "status" of undefined'

### DIFF
--- a/src/services/api.js
+++ b/src/services/api.js
@@ -15,7 +15,7 @@ export const initializeApi = ({ onUnauthorized }) => ({
     .then(response => {
       return response.data;
     })
-    .catch(({ response }) => {
+    .catch(response => {
       if (response.status === 401) {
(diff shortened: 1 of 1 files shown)


hlminh2000:
We'll keep an eye out on this one, but I think Axios treats error and success responses slightly differently...

hlminh2000:
maybe we can change the destructuring to happen inside the error handler body, and have a case for no response, so we can identify the issue more easily? 
The problem today was an infrastructure issue on the persona deployment, so doesn't seem there needs to be an application change, just to make it easier to spot the problem next time :P

jgnieuwhof:
@hlminh2000 - totally cool with closing this if it doesn't apply, however at some point we'll need to write some application code to gracefully handle these failed requests. It's a valid case that one of these services could go down in prod, and we don't want a runtime error because of it.

PR #329, opened 2018-05-02
🐛 allow anonymous portal user: UserSnap screen creation

bdolly:
adds a conditional for the widget email field population


src/services/usersnap.js
```diff
@@ -43,10 +43,15 @@ window._usersnapconfig = {
...
+      if (additionalUsersnapState.email) {
+        window.UserSnap.setEmailBox(additionalUsersnapState.email);
+      }
+    });
   },
 };
 
-export const init = () => scriptjs(urlJoin(usersnapHost, `${usersnapId}.js`));
+export const init = () => {
+  debugger;
```
jgnieuwhof: :fire:

PR #336, opened 2018-05-03
User Analytics (GA) tracking implementation

bdolly:
Implements analyticsTracking (GA) service to track user user actions and metrics. 

Tracks: 
• Join process
• Profile actions
• pageview
• outbound links


public/index.html
```diff
@@ -2,13 +2,13 @@
...
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-87708930-3"></script>
+ <!--    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-87708930-5"></script>
     <script>
       window.dataLayer = window.dataLayer || [];
       function gtag() { dataLayer.push(arguments); }
       gtag('js', new Date());
-      gtag('config', 'UA-87708930-3');
-    </script>
+      gtag('config', 'UA-87708930-5');
+    </script> -->
```
jgnieuwhof: 🔥

src/services/analyticsTracking.js
```diff
@@ -0,0 +1,158 @@
...
+        ...options,
+    });
+    ReactGA.pageview(page);
+};
+
+export const trackExternalLink = url => {
+    ReactGA.outboundLink({ label: url },() => {} );
+}
+
+export function withPageViewTracker(WrappedComponent, options = {}) {
```
jgnieuwhof: good intuition on the HOC, but I think there may be a simpler approach to this one.

`services/history` exports the history object that react router uses, check out documentation on `listen` here: https://github.com/ReactTraining/history.

something like 
```
  history.listen((location, action) => trackPageView(...))
```

when using ^^ you may want to whitelist paths, unless you want to track everything

src/services/profiles.js
```diff
@@ -121,7 +121,7 @@ export const deleteProfile = api => async ({ user }) => {
       `,
     },
   });
-
+
```
jgnieuwhof: Have this PR's changes been run through prettier?

src/uikit/ExternalLink.js
```diff
@@ -1,18 +1,24 @@
...
   return (
     <a
       {...props}
+      href={href}
       target="_blank"
       rel="noopener noreferrer"
       className={className + ' test-external-link'}
+      onClick={() => {
+        trackExternalLink(href);
+      }}
```
jgnieuwhof: really minor, but prefer inline handlers: `onClick={() => trackExternalLink(href)}`

src/uikit/Wizard.js
```diff
@@ -79,9 +80,12 @@ export default compose(
       setIndex(index + 1 >= steps.length ? index : index + 1),
     prevStep: ({ index, setIndex, steps }) => event => setIndex(index - 1 < 0 ? index : index - 1),
   }),
-  withPropsOnChange(['index'], ({ index, setIndex, steps }) => ({
-    currentStep: steps[index] || { title: 'no step', Component: '--' },
-  })),
+  withPropsOnChange(['index'], ({ index, setIndex, steps }) => {
```
jgnieuwhof: `react-router` has `withRouter`, which populates props w/ `{ match, location, history }`, which is preferred over `window.location`

src/App.js
```diff
@@ -85,7 +86,7 @@ const App = compose(injectState, withApi)(({ editing, setEditing, state, api })
             forceSelectRole({
               api,
               isLoadingUser,
-              Component: UserProfile,
+              Component: withPageViewTracker(UserProfile),
```
jgnieuwhof: view comment @ bottom re: HOC, I think we may be able to avoid the HOC altogether

src/components/FileRepo.js
```diff
@@ -31,6 +31,7 @@ import { arrangerProjectId } from 'common/injectGlobals';
 import Select from '../uikit/Select';
 import translateSQONValue from 'common/translateSQONValue';
 import { withApi } from 'services/api';
+import { withPageViewTracker } from 'services/analyticsTracking';
```
jgnieuwhof: "

src/components/Join.js
```diff
@@ -118,6 +119,10 @@ const Consent = compose(
               const { email, percentageFilled, ...rest } = loggedInUser;
               if (event.target.checked) {
                 setCustomStepMessage(null);
+                trackUserInteraction({
+                  category: 'Join',
```
jgnieuwhof: I wonder if we should export an object `constants` from `services/analyticsTracking`, that has categories / actions so we're not using magic strings when we track interactions.
```
import { constants } from 'services/analyticsTracking'
...
trackUserInteraction({
  category: constants.categories.join,
  action: constants.actions.acceptedTerms
});
```
?

src/components/Join.js
```diff
@@ -159,148 +164,157 @@ export const ButtonsDiv = styled('div')`
...
-                  />
-                  <ButtonsDiv>
+const JoinContent = compose(
+  injectState,
+  withRouter,
+  withTheme,
+  withApi,
+  lifecycle({
+    componentDidMount() {
+      startAnalyticsTiming('join process')
```
jgnieuwhof: same here w/ magic string

src/components/UserProfile/AboutMe.js
```diff
@@ -1,18 +1,29 @@
...
+import { compose, withState, withPropsOnChange, withHandlers } from 'recompose';
 import { withRouter } from 'react-router-dom';
 import { injectState } from 'freactal';
 import SaveIcon from 'react-icons/lib/md/save';
 import Autocomplete from 'react-autocomplete';
 import styled from 'react-emotion';
 
 import { withTheme } from 'emotion-theming';
-import EditableLabel from 'uikit/EditableLabel';
+import EditableLabel from 'uikit/EditableLabel';
```
jgnieuwhof: prettier?

src/components/UserProfile/AboutMe.js
```diff
@@ -43,6 +54,27 @@ export default compose(
   withState('isEditingBackgroundInfo', 'setEditingBackgroundInfo', false),
   withState('focusedTextArea', 'setFocusedTextArea', 'myBio'),
   withState('editingResearchInterests', 'setEditingResearchInterests', false),
+  withHandlers({
+    isEditingInfo: ({ setEditingBackgroundInfo, setEditingResearchInterests }) => ({
```
jgnieuwhof: here I think I'd prefer to just see two handlers, since they're basically two separate actions anyway. something like:
```
const trackProfileInteraction = ({ action, value }) => 
  trackUserInteraction({
    category: `User: Profile`,
    action: `Edit: ${action}`,
    label: value ? `open` : `close`
  });
...
withHandlers({
  handleEditingBackgroundInfo: ({ setEditingBackgroundInfo }) => ({
    type,
    value,
  }) => {
    setEditingBackgroundInfo(value);
    trackProfileInteraction({ action: 'Background Info', value });
  },
  // ... (repeat for research interests) 
})
```

src/components/UserProfile/AboutMe.js
```diff
@@ -120,7 +154,11 @@ export default compose(
             Background Information
             {canEdit &&
               (!isEditingBackgroundInfo ? (
-                <EditButton onClick={() => setEditingBackgroundInfo(!isEditingBackgroundInfo)} />
+                <EditButton
+                  onClick={() =>
+                    isEditingInfo({ type: 'background', value: !isEditingBackgroundInfo })
```
jgnieuwhof: use separate handlers in all these events

src/components/UserProfile/Gen3Connection.js
```diff
@@ -94,9 +106,9 @@ const Gen3Connection = ({
             <span className="numberBullet">1</span>
             <span>
               You will need to retrieve your authentication token from the{' '}
-              <a href={gen3WebRoot} rel="noopener noreferrer" target="_blank">
+              <ExternalLink href={gen3WebRoot} hasExternalIcon={false}>
```
jgnieuwhof: noice

src/components/forms/BasicInfoForm.js
```diff
@@ -112,6 +116,14 @@ export default compose(
         setErrors,
       }: any,
     ) => {
+      if (window.location.pathname.includes('/user') && values.roles !== loggedInUser.roles[0]) {
```
jgnieuwhof: `withRouter` for location in props

src/components/forms/SelectRoleForm.js
```diff
@@ -79,6 +83,10 @@ export const enhance = compose(
       }).then(
         async profile => {
           await setUser({ ...profile, email, api });
+          let { pathname } = window.location;
+          if (pathname === '/join') {
```
jgnieuwhof: `withRouter`

src/services/analyticsTracking.js
```diff
@@ -0,0 +1,158 @@
+import React, { Component } from 'react';
+import ReactGA from 'react-ga';
+import { addInfo as addUserSnapInfo } from './usersnap';
+import { gaTrackingID } from 'common/injectGlobals';
+import { merge } from 'lodash';
+
+const debug = process.env.NODE_ENV === 'development';
```
jgnieuwhof: this may be a useful constant to export from `common/injectGlobals`

src/services/analyticsTracking.js
```diff
@@ -0,0 +1,158 @@
...
+
+export const initAnalyticsTracking = () => {
+    ReactGA.initialize(GAState.trackingId, { debug });
+    ReactGA.ga(function(tracker) {
+        var clientId = tracker.get('clientId');
+        addStateInfo({ clientId });
+        addUserSnapInfo({ gaClientId: clientId });
+        ReactGA.set({ clientId: GAState.clientId });
+    });
+};
```
jgnieuwhof: This file doesn't look prettier'd?

jgnieuwhof: also, consistently using `const` arrow functions is preferred

src/services/analyticsTracking.js
```diff
@@ -0,0 +1,158 @@
...
+
+const debug = process.env.NODE_ENV === 'development';
+
+let GAState = {
+    trackingId: gaTrackingID,
+    userId: null,
+    userRoles: null,
+    clientId: null,
+};
+let timingsStorage = window.localStorage;
```
jgnieuwhof: Do we actually need localStorage here? If we can get away with a singleton object that's held in memory I'd be happy with that, unless the business requirement entails LS.

src/services/profiles.js
bdolly: I have it set to run on save command in Sublime Text, I may missed a keystroke when saving some of the files. I think it's possible to write a git-hook so it all commits or PRs get ran through prettier isn't it?

src/services/analyticsTracking.js
bdolly: yeah was wondering if there was a better approach, feel down a few holes because the HOC caused re-rendering issues

src/components/Join.js
bdolly: This is definitely on my todo list. Was just trying to squeeze some of the out before internal review so we can iterate on it once we go into ALPHA/BETA.

src/services/analyticsTracking.js
bdolly: idea here is that localStorage is persisted between sessions. That way we could possibly track super long timings and use timing value outliers as possible indications for areas of confusion

src/services/profiles.js
jgnieuwhof: Yeah I think you're right

src/services/analyticsTracking.js
jgnieuwhof: Yeah that's fair, cool cool.

src/components/UserProfile/AboutMe.js
```diff
@@ -212,19 +239,21 @@ export default compose(
...
-                  setEditingBackgroundInfo(false);
+                  handleEditingBackgroundInfo({ value: false });
                 }}
                 css={theme.hollowButton}
               >
                 Cancel
               </button>
               <SaveButton
                 onClick={async () => {
+                  debugger;
```
jgnieuwhof: :fire:

jgnieuwhof:
hey @bdolly - is this still WIP at this point? noticed some debugging statements in the code that made me think this might be the case

PR #337, opened 2018-05-03
🐛 fix user create profile 

jgnieuwhof:
createProfile was broke, data is not nested under data no more

### DIFF
--- a/src/services/profiles.js
+++ b/src/services/profiles.js
@@ -29,7 +29,9 @@ const DEFAULT_FIELDS = `
 `;
 
 export const getProfile = api => async ({ egoId }) => {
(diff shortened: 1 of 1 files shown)


src/services/profiles.js
```diff
@@ -54,7 +56,11 @@ export const getProfile = api => async ({ egoId }) => {
 };
 
 export const createProfile = api => async ({ egoId, lastName, firstName, email }) => {
-  const { data: { data: { userCreate: { record } } } } = await api({
+  const {
+    data: {
+      userCreate: { record },
+    },
+  } = await api({
```
jgnieuwhof: this is the one, the rest was just prettier

PR #343, opened 2018-05-07
✨ harmonize clinical reports with file reports

jgnieuwhof:
a couple things:
- extracted dropdown w/ loading option into its own component, separating it from the file manifest business logic
- file manifest dropdown & clinical file dropdown both use new ^^ extracted component
- participant / participant + family clinical file changes (columns, sort, uniqueBy removed)


src/uikit/PillInputWithLoadingOptionsAndButton.js
```diff
@@ -0,0 +1,138 @@
...
+      width: 15,
+      height: 15,
+      marginLeft: 7,
+      position: 'absolute',
+    }}
+  />
+);
+
+class PillInputWithLoadingOptionsAndButton extends React.Component {
+  initOptionState = (props, val = { enabled: true }) => mapValues(props.options, () => val);
```
hlminh2000: just nit picking :P kinda wanna see `loading: false` in val

jgnieuwhof: yeah good call :)

PR #344, opened 2018-05-07
🐛 api missing in profile delete button

hlminh2000:


hlminh2000:
the DeleteButton component is used for canceling during signup

PR #345, opened 2018-05-07
🐛 passes a ref that allows term agg to stay in view on 'less'

jgnieuwhof:
depends on https://github.com/overture-stack/arranger/pull/363

### DIFF
--- a/src/components/FileRepo.js
+++ b/src/components/FileRepo.js
@@ -125,9 +125,11 @@ const UploadIdsButton = ({ theme, state, effects, setSQON, ...props }) => (
   </div>
 );
 
+const aggregationsWrapperRef = React.createRef();
 const AggregationsWrapper = enhance(({ state, effects, theme, setSQON, ...props }) => {
   return (
     <div
+      ref={aggregationsWrapperRef}
       css={`
         height: 100%;
         height: calc(100vh - 180px);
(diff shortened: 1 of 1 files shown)


src/components/FileRepo.js
```diff
@@ -199,7 +201,7 @@ const AggregationsWrapper = enhance(({ state, effects, theme, setSQON, ...props
         />
         <UploadIdsButton {...{ theme, effects, state, setSQON, ...props }} />
       </div>
-      <Aggregations {...{ ...props, setSQON }} />
+      <Aggregations {...{ ...props, setSQON, containerRef: aggregationsWrapperRef }} />
```
hlminh2000: YES!!! React 16.3 rocks!!!!

src/components/FileRepo.js
```diff
@@ -125,9 +125,11 @@ const UploadIdsButton = ({ theme, state, effects, setSQON, ...props }) => (
   </div>
 );
 
+const aggregationsWrapperRef = React.createRef();
```
hlminh2000: YES!!! React 16.3 rocks!!!!

hlminh2000: woops! maybe declare this inside `AggregationsWrapper` so we don't have a singleton ref :P

jgnieuwhof:
@hlminh2000 - addressed the singleton issue :)

PR #348, opened 2018-05-07
✨ update gen3 url, use .env rather than hardcoded value

jgnieuwhof:
- adds gen3 web url to .env
- need to add `REACT_APP_GEN3_WEB` to netlify env vars before deploying


hlminh2000:
just commenting to satisfy my OCD and mark this as reviewed :P

PR #349, opened 2018-05-07
💄 updated profile banners, colors, & icons

jgnieuwhof:
~~[WIP] until icons are updated as well (waiting on those)...~~

Changes:
- banner gradient now shared between profile info bar & user profile header
- updated background images for user profile header
- icons updated to most recent version


hlminh2000:
seems the color didn't change, why the preference for rgb? hex seems to be the standard in the code so just wondering.

jgnieuwhof:
@hlminh2000 - relatively thoughtless change, they came that way from the design so I left it, will revert the change if they're the same lol

jgnieuwhof:
@hlminh2000 - addressed

PR #351, opened 2018-05-08
🐛 enable bool & date fields in advanced facet view

jgnieuwhof:
depends on https://github.com/overture-stack/arranger/pull/366

arranger now defaults hiding id & text, which is what we need in KF


src/components/AdvancedFacetViewModal/index.js
```diff
@@ -37,14 +37,9 @@ class AdvancedFacetViewModalContent extends React.Component {
         >
           <div style={{ position: 'relative', flex: 1 }}>
             <AdvancedFacetView
-              {...{
-                ...props,
-                sqon: localSqon,
-                fieldTypesToExclude: ['id', 'text', 'date', 'boolean'],
```
hlminh2000: lololol I remember this now :P was a quick pre-demo thing

PR #367, opened 2018-05-10
🐛 accurately udpate CompleteOMeter per user role

bdolly:
accounts for some user roles not having all default fields. 

Use Case: 
if you created an account with the Community Role and filled out all available fields on the screen the CompleteOMeter would not show 100% because it accounted for all fields for all roles

### DIFF
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -4,6 +4,29 @@ import HealthIcon from 'icons/HealthIcon';
 import HealthBanner from 'assets/user-banner-health.png';
 import ResearchBanner from 'assets/user-banner-research.png';
 import CommunityBanner from 'assets/user-banner-community.png';
+import { without } from 'lodash';
+
+const DEFAULT_PROFILE_FIELDS = [
+      '_id',
+      'title',
+      'firstName',
+      'lastName',
+      'egoId',
+      'roles',
+      'acceptedTerms',
+      'email',
+      'story',
+      'bio',
+      'jobTitle',
+      'institution',
+      'city',
+      'state',
+      'country',
+      'website',
+      'googleScholarId',
+      'interests',
+      'sets',
+    ]
 
 export const ROLES = [
   {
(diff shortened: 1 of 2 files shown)


src/common/constants.js
```diff
@@ -4,6 +4,29 @@ import HealthIcon from 'icons/HealthIcon';
 import HealthBanner from 'assets/user-banner-health.png';
 import ResearchBanner from 'assets/user-banner-research.png';
 import CommunityBanner from 'assets/user-banner-community.png';
+import { without } from 'lodash';
+
+const DEFAULT_PROFILE_FIELDS = [
```
jgnieuwhof: I'm not sure this is actually necessary here. It may be misleading as it's disconnected from the actual form rendering (it only has an impact on percentage filled). The UI does an exclusion of `jobTitle` & `institution` in the jsx, I'd be more apt to follow that lead and just do the same exclusion in the `setUser` side effect

src/stateProviders/provideLoggedInUser.js
```diff
@@ -48,11 +49,16 @@ export default provideState({
...
         .then(({ data }) => {
-          return get(data, '__type.fields', []).length;
+          return get(data, '__type.fields', []);
         })
-        .then(totalFields => state => {
-          const filledFields = Object.values(user || {}).filter(v => v || (isArray(v) && v.length));
-          const percentageFilled = filledFields.length / totalFields;
+        .then(fields => state => {
+          const userRole = user.roles ? user.roles[0] : null;
+          const userRoleProfileFields = userRole ?  find(ROLES, { type: userRole }).profileFields : fields;
```
jgnieuwhof: as per the message above, without the ROLES thing this would be
`(userRole && userRole !== 'researcher') ? without(fields, 'institution', 'jobTitle') : fields`

src/stateProviders/provideLoggedInUser.js
```diff
@@ -48,11 +49,16 @@ export default provideState({
...
+          const profile = pick(user, userRoleProfileFields);
+          const filledFields = Object.values(profile || {}).filter(
+            v => v || (isArray(v) && v.length),
```
jgnieuwhof: not sure if I understand this filter, afaik we'll never hit the fallback because an empty array is truthy

PR #378, opened 2018-05-15
Analytics Tracking for File Repo actions

bdolly:
TRACKS:
- Filter ( and sqon) changes
- DataTable 
- - Export TSV 
- - Text Filter
- - Share Query (by channel)
-- Save Query
-- All Actions in the right sidebar

NOTE: some of the on<action> prop calls from Arranger are dependent on Arranger PRs but rolling them in here to catch those events once merged to Arranger


src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -127,6 +128,38 @@ export default compose(
       }).then(x => this.setState(x));
     },
   }),
+  withFormik({
```
jgnieuwhof: this seems odd, there's already a similar form down below in the body doing a similar thing w/ `fileManifestParticipantsAndFamily`, but as far as i can tell this one wouldn't be used. is this tested?

src/components/FileRepo.js
```diff
@@ -31,9 +32,25 @@ import { arrangerProjectId } from 'common/injectGlobals';
...
 import translateSQONValue from 'common/translateSQONValue';
 import { withApi } from 'services/api';
+import { trackUserInteraction, TRACKING_EVENTS } from 'services/analyticsTracking';
 import ArrangerConnectionGuard from './ArrangerConnectionGuard';
 import { config as statsConfig } from './Stats';
 
+const trackFileRepoInteraction = eventData => {
+  if (eventData.label && isObject(eventData.label)) {
+    eventData.label = JSON.stringify(eventData.label);
+  }
```
jgnieuwhof: prefer non-mutation, so could do
```
const trackFileRepoInteraction = ({ label, ...eventData }) =>
  trackUserInteraction({
    ...eventData,
    ...(label && { label: isObject(label) ? JSON.stringify(label) : label }),
    category: 'File Repo',
    action: 'default file repo action'
  })
```

src/components/FileRepo.js
```diff
@@ -168,13 +185,43 @@ const AggregationsWrapper = compose(injectState, withTheme)(
...
-                  {...props}
-                  {...{ statsConfig }}
-                  closeModal={effects.unsetModal}
-                  onSqonSubmit={({ sqon }) => {
-                    setSQON(sqon);
-                    effects.unsetModal();
+                  {...{
+                    ...props,
+                    closeModal: effects.unsetModal,
+                    onClear: () => {
```
jgnieuwhof: onClear, onFilterChange, onTermSelected, weren't previously in `props` ? just want to make sure we're not overwriting these handlers that may have been passed down from parent components

src/components/FileRepoSidebar/ReportsDownloadInput.js
```diff
@@ -36,7 +37,14 @@ export default ({ api, projectId, theme, sqon }) => (
                   },
                 },
                 columns: state.columns,
-              })();
+              })().then(downloadData => {
```
jgnieuwhof: would it make sense to move these events into the file download helper functions themselves? that way regardless of where / when they're called in the UI the events still get tracked the same

src/components/ParticipantManifestModal.js
```diff
@@ -24,7 +25,22 @@ const ParticipantManifestModal = ({
           projectId,
           setWarning,
           onDownloadClick: async () => {
-            await fileManifestParticipantsOnly({ sqon, columns })();
+            await fileManifestParticipantsOnly({ sqon, columns })()
```
jgnieuwhof: same q here re: moving into download function itself

src/components/ShareSaveQuery/SaveQuery.js
```diff
@@ -48,9 +51,20 @@ export default injectState(
             loading: false,
             link: urlJoin(arrangerApiRoot, 's', data.id),
           });
+          sqon.id = data.id;
```
jgnieuwhof: shouldn't mutate sqon here, would prefer a copy: `const trackingSqon = { ...sqon, id: data.id };`

src/components/cavatica/CavaticaCopyModal.js
```diff
@@ -208,9 +209,19 @@ const CavaticaCopyModal = ({
...
+                action: 'Copied Files to Cavatica Project',
+                label: uuids,
+              });
               props.onComplete();
             } catch (e) {
               //TODO: Display failure error.
+              trackUserInteraction({
+                category: TRACKING_EVENTS.categories.fileRepo.actionsSidebar,
+                action: 'Copied Files to Cavatica Project FAILED',
+                label: e,
```
jgnieuwhof: `e.message`, or stack trace? not sure if we want to jam in the whole error object here

src/components/FileRepo.js
jgnieuwhof: on this note, I've seen the `JSON.stringify` label thing a couple times. We could do this same `label` handling business in the `trackUserInteraction` function so it doesn't have to be repeated by callers

src/components/FamilyManifestModal/FamilyManifestModal.js
hlminh2000: this looks like the old code, which was removed during the reimplementation of the modal... probably an artifact of merge conflict resolution...

src/components/ParticipantManifestModal.js
hlminh2000: I believe this file is no longer referenced as both the two download modal have been merged to one

src/components/FileRepo.js
bdolly: double checked, these are all props that don't current exist on AFV. I put these in this PR so that when these events get merged to Arranger the file repo tracking will be all set.

src/components/FileRepoSidebar/ReportsDownloadInput.js
bdolly: wanted to try and keep these as reasonably close to the display layer as possible, since they are tracking user events more so than programatic events. Moving them to the service layer would require passing contextual information in args which felt like crossing the separation of concerns.

src/components/FileRepo.js
jgnieuwhof: 👍

src/components/FileRepoSidebar/ReportsDownloadInput.js
jgnieuwhof: fair :)

src/components/ParticipantManifestModal.js
jgnieuwhof: oh yeah true, was this a merge mishap? or was it there / we just forgot to remove it?

bdolly: I think this was a merge mishap, I saw it and didn't want to remove because I hadn't looked at Minh's modal updates

PR #384, opened 2018-05-16
🐛 standardize translate SQON, pass to AFV for sqon view

jgnieuwhof:
2 things here:
- standardizes our initialization of translateSQON
- passes translateSQON to AFV 'cause it now accepts it too

depends on https://github.com/overture-stack/arranger/pull/390


src/components/FileRepo.js
```diff
@@ -180,6 +188,7 @@ const AggregationsWrapper = compose(injectState, withTheme)(
                 <AdvancedFacetViewModalContent
                   {...{
                     ...props,
+                    translateSQONValue,
```
hlminh2000: technically this and the `translateSQONValue` below would be in `props` before so this extraction isn't necessary, but if it's here for readability, that's fine for me too :)

PR #386, opened 2018-05-16
removes some dom element and sets modal maxHeight

hlminh2000:
requires https://github.com/overture-stack/arranger/pull/391

### DIFF
--- a/src/components/Modal/index.js
+++ b/src/components/Modal/index.js
@@ -197,6 +197,12 @@ const ModalView = ({
   >
     {!!title ? <ModalHeader {...{ theme, title, unsetModal, ...props }} /> : null}
     <div
+      className={`${css`
+        z-index: 1000;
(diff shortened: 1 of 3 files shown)


src/components/Modal/index.js
```diff
@@ -197,10 +198,14 @@ const ModalView = ({
   >
     {!!title ? <ModalHeader {...{ theme, title, unsetModal, ...props }} /> : null}
     <div
-      css={`
+      className={`${css`
         z-index: 1000;
         flex: 1;
+        display: flex;
+        flex-direction: column;
```
jgnieuwhof: this is a safe global change for the other modals?

hlminh2000: I went through and checked the other modals which looked fine but you're right, probably safer to have this set specifically for this

PR #397, opened 2018-05-22
💄 make role select form scrollable

hlminh2000:
pretty much the same strategy as the modal :)


hlminh2000:
note to self: hold off merging till end of freeze

PR #398, opened 2018-05-22
👌 improvement to manifest download modal UI

hlminh2000:
addressing these points from Rosi:
1) When the family manifest loads, there is a weird bump as it calculates whether or not to add family... Can we instead calculate before the modal loads and then load the correct size modal?

2) When you click Generate, there is a flash before it switches to copy. Can you use the loader instead to prevent the weird flash?

3) Add a bit more space between Download button and Generate manifest button (like screenshot above)

### DIFF
--- a/src/components/FamilyManifestModal/FamilyManifestModal.js
+++ b/src/components/FamilyManifestModal/FamilyManifestModal.js
@@ -121,7 +121,7 @@ export default compose(
   injectState,
   withApi,
   lifecycle({
-    state: { loadingProps: true },
+    state: { loading: true },
     componentDidMount() {
       generateFamilyManifestModalProps({
         api: this.props.api,
@@ -129,7 +129,7 @@ export default compose(
       }).then(x =>
         this.setState({
           ...x,
-          loadingProps: false,
+          loading: false,
         }),
       );
     },
@@ -139,7 +139,7 @@ export default compose(
   withState('checkedFileTypes', 'setCheckedFileTypes', []),
 )(
   ({
-    loadingProps,
+    loading,
     theme,
     familyMemberIds,
     participantIds,
@@ -244,7 +244,7 @@ export default compose(
               </Table>
             </Section>
           );
-          return loadingProps ? (
+          return loading ? (
             spinner
           ) : (
             <div className={`${theme.column} ${modalContentStyle(theme)}`}>
@@ -256,7 +256,7 @@ export default compose(
                     participantIds,
                     projectId,
                     isDisabled,
-                    render: ({ loading, fileTypeStats }) => {
+                    render: ({ loading: loadingFileTypeStats, fileTypeStats }) => {
                       const uniqueParticipantsAndFamilyMemberIds = uniq([
                         ...participantIds,
                         ...(fileTypeStats
@@ -269,7 +269,7 @@ export default compose(
                       const FooterWithParticipantsAndFamilyMembers = createFooterComponent(
                         uniqueParticipantsAndFamilyMemberIds,
                       );
-                      return loading ? (
+                      return loadingFileTypeStats ? (
                         spinner
                       ) : (
                         <Fragment>


src/components/DownloadManifestModal.js
```diff
@@ -88,13 +95,16 @@ const GenerateManifestSet = compose(injectState, withTheme)(
...
-              });
+              const [{ data, errors }] = await Promise.all([
+                saveSet({
+                  type,
+                  sqon: sqon || {},
+                  userId: loggedInUser.egoId,
+                  path: 'kf_id',
+                  api: graphql(api),
+                }),
+                wait(1000),
```
hlminh2000: it seems the request tends to be so fast that it looks like a flash so we're putting in an arbitrary delay

src/components/DownloadManifestModal.js
```diff
@@ -16,6 +16,13 @@ import LoadingOnClick from 'components/LoadingOnClick';
 import graphql from '../services/arranger';
 import Spinner from 'react-spinkit';
 
+const wait = async ms =>
+  new Promise((resolve, reject) => {
+    setTimeout(() => {
+      resolve('done');
+    }, ms);
+  });
```
jgnieuwhof: not `async` since we're not using await. can be simplified to:
```
  const wait = (s = 1) => new Promise(r => setTimeout(r, s * 1000))
```

src/components/FamilyManifestModal/FamilyDataTypesStatsQuery.js
```diff
@@ -60,8 +60,8 @@ export default compose(withApi)(
       }, {})}
       render={({ data, loading }) => {
         return render(
-          loading
-            ? { loading, data: [] }
+          loading || !data
+            ? { loading: true, data: [] }
```
jgnieuwhof: this seems odd.. if this `loading` prop is driving a spinner this would lead to a seemingly infinite load, my guess is that that's not actually what we want?

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -257,10 +260,11 @@ export default compose(
                       const FooterWithParticipantsAndFamilyMembers = createFooterComponent(
                         uniqueParticipantsAndFamilyMemberIds,
                       );
-                      return loading ? (
+                      return loading || !fileTypeStats.length ? (
```
jgnieuwhof: this seems a little odd too, is a lack of length here actually a loading state? what if there are legitimately no file type stats?

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -207,38 +207,41 @@ export default compose(
...
+                    files: participantFilesCount,
+                    fileSize: fileSizeToString(participantFilesSize),
+                    isChecked: isFamilyMemberFilesAvailable,
+                    leftComponent: <CheckCircleIcon className={`checkMark`} />,
+                  }}
+                />
+              </Table>
+            </Section>
+          );
+          return !dataTypes ? (
```
jgnieuwhof: shouldn't this be driven by a loading state? what if there are no data types?

hlminh2000: the empty state would come as `[]`, but for better clarity, I've added `loadingProps` flag :)

hlminh2000:
Nothing should technically break the way it was, but it was definitely difficult to reason out why so I've cleaned it up but we need this for it to work :) https://github.com/overture-stack/arranger/pull/399

jgnieuwhof:
yeah good catch on fixing the initialization of that loading state! keeping things clean on this side is 👌

src/components/FamilyManifestModal/queries.js
```diff
@@ -106,15 +106,16 @@ export const generateFamilyManifestModalProps = async ({ api, sqon }) => {
...
   const { participantFilesCount, participantFilesSize } = await participantsFilesCountAndSize({
     api,
     sqon,
   });
+  const dataTypes = await dataTypesFromFamilyMemberIds({
+    api,
+    familyMemberIds: familyMembersWithoutParticipantIds,
+  });
   return {
+    loadingProps: false,
```
jgnieuwhof: what's this `loadingProps`? from what I can tell it is only used once and will always be false?

src/components/DownloadManifestModal.js
```diff
@@ -16,6 +16,8 @@ import LoadingOnClick from 'components/LoadingOnClick';
 import graphql from '../services/arranger';
 import Spinner from 'react-spinkit';
 
+const wait = (s = 1) => new Promise(r => setTimeout(r, s));
```
jgnieuwhof: `s * 1000` I think 😛

hlminh2000: oops :P good catch :)

src/components/FamilyManifestModal/queries.js
hlminh2000: it's `true` when the component loads and becomes `false` when `generateFamilyManifestModalProps` resolves.

hlminh2000: oh yeah I somehow missed that `generateFamilyManifestModalProps` was already in a life cycle hook :P yeah that'd make it easier to read definitely :)

src/components/FamilyManifestModal/queries.js
```diff
@@ -106,14 +106,14 @@ export const generateFamilyManifestModalProps = async ({ api, sqon }) => {
...
-    familyMemberIds: familyMembersWithoutParticipantIds,
-  });
   const { participantFilesCount, participantFilesSize } = await participantsFilesCountAndSize({
     api,
     sqon,
   });
+  const dataTypes = await dataTypesFromFamilyMemberIds({
+    api,
+    familyMemberIds: familyMembersWithoutParticipantIds,
+  });
```
jgnieuwhof: Should we do a Promise all here since there's no dependency? 
```
const [{ participantFilesCount, participantFilesSize }, dataTypes] = await Promise.all([
  participantsFilesCountAndSize(...),
  dataTypesFromFamilyMemberIds(...),
]);
```

hlminh2000: `dataTypes` actually relies on `familyMembersWithoutParticipantIds` from `familyMemberAndParticipantIds` :) but definitely makes sense for the first two. Addressed! :)

src/components/FamilyManifestModal/FamilyManifestModal.js
```diff
@@ -121,18 +121,25 @@ export default compose(
   injectState,
   withApi,
   lifecycle({
+    state: { loadingProps: true },
```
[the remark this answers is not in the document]
hlminh2000: addressed! made the other `loading` a more fancy `loadingFileTypeStats` :)

PR #399, opened 2018-05-22
✨ added auto-complete interests from persona

jgnieuwhof:
persona supports interests searching now - this uses that to provide a list of suggestions when entering interests. users can add their own interests by entering a new one.

### DIFF
--- a/src/components/UserProfile/InterestsAutocomplete.js
+++ b/src/components/UserProfile/InterestsAutocomplete.js
@@ -9,31 +9,58 @@ import { Trans } from 'react-i18next';
 import { withApi } from 'services/api';
 import { getTags } from 'services/profiles';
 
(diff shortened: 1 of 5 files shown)


src/components/UserProfile/InterestsAutocomplete.js
```diff
@@ -0,0 +1,162 @@
...
+import Downshift from 'downshift';
+import { Trans } from 'react-i18next';
+
+import { withApi } from 'services/api';
+import { getTags } from 'services/profiles';
+
+const DropdownLabel = compose(withTheme)(({ theme, children }) => (
+  <span
+    className={css`
+      text-transform: uppercase;
```
hlminh2000: I think we should start trying to move new styles outside of components, maybe just in a local style js file so it's easier to move things around later :)

PR #406, opened 2018-05-23
⬆️ updates arranger components

hlminh2000:


hlminh2000:
note to self: hold off till end of freeze

PR #414, opened 2018-05-24
✨ sliding action panel + style updates

hlminh2000:


src/theme/defaultTheme.js
```diff
@@ -32,6 +32,11 @@ const colors = {
 };
 
 const components = {
+  globalStyleOverride: css`
+    & button {
+      border: solid 1px ${colors.borderGrey};
+    }
+  `,
```
jgnieuwhof: not sure about the intent here.. are we trying to globally override other styles that we've used in child components? if that's the case it seems more appropriate to fix the things that we're trying to override, rather than overriding them. but maybe I'm missing the point? thoughts?

hlminh2000: it's there to override browser defaults, which oddly enough look different in Chrome on windows vs Chrome on Mac. If any styling issue arises I think it can be addressed on a case-by-case basis

jgnieuwhof: oh gotcha, yeah sounds good

PR #418, opened 2018-05-25
✨ download mechanism now triggered by click on icon in row

jgnieuwhof:
a couple things:
- removed download button from actions sidebar
- added a download column to the table
- cleaned up how some components were getting their props (make things more self-contained)
- standardized the download icon into its own component


jgnieuwhof:
depends on https://github.com/overture-stack/arranger/pull/403

PR #423, opened 2018-05-28
💄 Style/account screen

hlminh2000:


src/uikit/IntegrationStepsModal.js
```diff
@@ -0,0 +1,57 @@
+import styled from 'react-emotion';
+import { withTheme } from 'emotion-theming';
+
+export const NumberBullet = styled.span`
```
jgnieuwhof: I _think_ `uikit` things are supposed to be generic components that are used across the app, `IntegrationStepsModal` seems relatively specific. If these ones are generic (can be used in modals or pages other than the two integration ones) then we may want to rename them to something less specific, or if they are in fact specific then I'd think this component may just belong in `Components`. I dunno, just trying to think about how to keep things organized / promote future re-use, what do you think?

hlminh2000: yeah I think it'll make sense in Component... Thanks! :D

PR #425, opened 2018-05-28
🐛 fix some modal navigation / functionality issues

jgnieuwhof:
A couple things:
- arranger update to fix upload ids modal
- fixes a race condition when setting new SQON on AFV close

### DIFF
--- a/src/components/FileRepo/FileRepo.js
+++ b/src/components/FileRepo/FileRepo.js
@@ -220,8 +220,7 @@ const AggregationsWrapper = compose(injectState, withTheme)(
                           }
                         },
                         onSqonSubmit: ({ sqon }) => {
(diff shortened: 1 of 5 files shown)


src/components/FileRepo/FileRepo.js
```diff
@@ -220,13 +220,13 @@ const AggregationsWrapper = compose(injectState, withTheme)(
                           }
                         },
                         onSqonSubmit: ({ sqon }) => {
-                          setSQON(sqon);
+                          effects.unsetModal();
+                          setTimeout(() => setSQON(sqon), 0);
```
hlminh2000: let's move this inside the unsetModal as a callback so we can minimize the use of this pattern :P (Jeff is watching!! :)))

PR #446, opened 2018-05-31
💄more css refactor + some stub-ins for mailing list subscription

jgnieuwhof:
I was working on mailing list subscription but will be doing import / export arranger config today. However I think there may be enough css changes here to warrant getting this in asap & continuing with the mailing list stuff later. I've commented out the mailing stuff w/ TODO notes for now until I can pick that up again.

### DIFF
--- a/src/components/Join.js
+++ b/src/components/Join.js
@@ -2,14 +2,13 @@ import React, { Fragment } from 'react';
 import { injectState } from 'freactal';
 import { withRouter } from 'react-router-dom';
 import { compose, withState, lifecycle } from 'recompose';
-import styled, { css } from 'react-emotion';
+import styled from 'react-emotion';
 import { withTheme } from 'emotion-theming';
 import { Trans } from 'react-i18next';
 import LeftIcon from 'react-icons/lib/fa/angle-left';
 import RightIcon from 'react-icons/lib/fa/angle-right';
-
 import { get } from 'lodash';
-import Wizard from 'uikit/Wizard';
+
 import Login from 'components/Login';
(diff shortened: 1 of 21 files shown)


src/components/Join.js
```diff
@@ -103,59 +104,38 @@ const Consent = compose(
...
-          </Trans>
-        </div>
-        {customStepMessage && (
-          <div
-            className={css`
-              color: red;
-              width: 90%;
-            `}
-          >
-            {customStepMessage}
```
hlminh2000: was this just never used? Just wanted to confirm it's safe to remove

jgnieuwhof: Yeah good q. Other steps of the wizard disable buttons if you haven't completed what you need to do to go to the next step. For some reason this step was leaving the button enabled and displaying a message that you need to accept terms in order to move forward. Now, just like the other steps we just disable the button until the thing is checked.

It seems more consistent, and the visual feedback is pretty obvious - initially the checkbox bubble is white and the button is disabled / greyed out, when you click the checkbox the checkbox bubble activates in addition to the button turning blue & being enabled

hlminh2000: uhm yeah that seems better, it kinda feels like a trap otherwise eh :P 
But then I feel like maybe there was a reason why we ended with this behaviour, thinking it might be better to confirm with the BA before we change it, just in case...

jgnieuwhof: yeah good call, @rosibaj - you have any thoughts on this?

rosibaj: this is a good point @hlminh2000 @jgnieuwhof .  I like your suggestions to keep the button disabled until the terms and conditions are select.

PR #451, opened 2018-06-04
✨ feeds custom FilterInput component to Arranger components

hlminh2000:
Added a FilterInput that extends the react-dom's input component for consistency.

Merge https://github.com/overture-stack/arranger/pull/413 first then I'll bundle the update in here

### DIFF
--- a/src/components/AdvancedFacetViewModal/index.js
+++ b/src/components/AdvancedFacetViewModal/index.js
@@ -55,7 +55,7 @@ class AdvancedFacetViewModalContent extends React.Component {
         <AfvContainer>
           <AdvancedFacetView
             {...props}
-            FilterInputComponent={props => (
+            InputComponent={props => (
               <div>
                 <FilterInput {...props} />
               </div>
--- a/src/components/AdvancedFacetViewModal/index.js
+++ b/src/components/AdvancedFacetViewModal/index.js
@@ -6,6 +6,7 @@ import { compose } from 'recompose';
 import { ModalFooter } from '../Modal/index.js';
 import { AdvancedFacetView } from '@arranger/components/dist/Arranger';
 import { provideLocalSqon } from 'stateProviders';
+import { FilterInput } from 'uikit/Input';
 
 const enhance = compose(provideLocalSqon, injectState);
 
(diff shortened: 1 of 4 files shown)


src/components/AdvancedFacetViewModal/index.js
```diff
@@ -54,6 +55,11 @@ class AdvancedFacetViewModalContent extends React.Component {
         <AfvContainer>
           <AdvancedFacetView
             {...props}
+            FilterInputComponent={props => (
+              <div>
```
jgnieuwhof: Is this <div> necessary? it looks like `FilterInput` is already wrapped in a `div`, and if it's not necessary then this could be `FilterInputComponent={FilterInput}`

src/uikit/Input.js
```diff
@@ -1,4 +1,47 @@
...
+  ({
+    theme,
+    LeftIcon = SearchIcon,
+    RightIcon = FaTimesCircleO,
+    className,
+    value,
+    disabled,
+    ...props
+  }) => {
+    const Wrapper = styled(Row)`
```
jgnieuwhof: should be declared outside this component so it's not re-instantiated every render

src/uikit/Input.js
```diff
@@ -1,4 +1,47 @@
+import React from 'react';
 import styled from 'react-emotion';
+import { withTheme } from 'emotion-theming';
+import Row from 'uikit/Row';
+import SearchIcon from 'react-icons/lib/fa/search';
+import FaTimesCircleO from 'react-icons/lib/fa/times-circle';
+
+// this this component should implement the default <input> type from react-dom with some additional
+// props for styling
+export const FilterInput = withTheme(
```
jgnieuwhof: w/ styled component no need for `withTheme` here

src/uikit/Input.js
```diff
@@ -1,4 +1,47 @@
...
+      }
+      .icon-right {
+        cursor: pointer;
+      }
+      input {
+        border: none;
+        flex: 1;
+      }
+    `;
+    const clearInput = () => props.onChange({ target: { value: '' } });
```
jgnieuwhof: I'm worried about this, seems more natural to place a ref on `input` and call `inputRef.current.value = ''` so that the browser can take care of triggering events

src/components/AdvancedFacetViewModal/index.js
hlminh2000: there was a bit of a styling issue that came from the theme's input style. Applied a better fix by overriding that style instead ;)

src/uikit/Input.js
hlminh2000: so this had to be a bit more involved than setting `inputRef.current.value`, but we have a pretty low-level browser-driven implementation now.

PR #453, opened 2018-06-04
✨ email subscriptions

hlminh2000:
needs https://github.com/kids-first/kf-persona/pull/1 first


src/components/forms/SelectRoleForm.js
```diff
@@ -252,8 +251,7 @@ export default enhance(
             </Column>
             {touched.roles && errors.roles && <div>{errors.roles}</div>}
           </Row>
-          {/* TODO: Issue #321 */}
-          {/* <Row buffer>
+          <Row buffer>
```
jgnieuwhof: should we nix `buffer` now that we can use `<Row mt={2}>` ? it's only a 2px difference, and means we're not using custom helpers that don't respect the theme spacing

src/components/Join.js
```diff
@@ -228,62 +104,9 @@ const JoinContent = compose(
...
-                    >
-                      <Trans>Save</Trans>
-                      <RightIcon />
-                    </button>
-                  </div>
-                </ButtonsDiv>
-              </Fragment>
+              <ConsentForm
+                {...{ disableNextStep, nextStep, prevStep, nextDisabled, prevDisabled, history }}
+              />
```
jgnieuwhof: 👌

src/components/forms/SelectRoleForm.js
hlminh2000: YES! will add this.

Issue #434, opened 2018-05-29
Fix Download File


rosibaj:
When i try and download a file, this 400 error is occurring:

![Screen Shot 2018-05-29 at 10.40.29 AM.png](https://images.zenhubusercontent.com/5a06111e8a75884b9086f5d8/6effaefa-725f-4860-9926-56007f2b8ae0)

```
{
  "errors": [
    {
      "message": "Cannot query field \"uuid\" on type \"fileNode\". Did you mean \"id\"?",
      "locations": [
        {
          "line": 1,
          "column": 79
        }
      ]
    }
  ]
}
```

I have access to this dataset so it should be downloadable - it should be querying on File ID.

hlminh2000:
Let's move this back to backlog as we're blocked by a data model update here it seems

Issue #430, opened 2018-05-28
Cavatica File push


rosibaj:
- [x] 1. Fix create project buttons - they should take up the whole space and be aligned
CURRENT:
![Screen Shot 2018-05-28 at 5.46.07 PM.png](https://images.zenhubusercontent.com/5a06111e8a75884b9086f5d8/5511400c-4314-44d0-8e07-044821fa1d67)
EXPECTED: 
![screen shot 2018-05-28 at 5 46 41 pm](https://user-images.githubusercontent.com/15824252/40630254-1f330ad0-629f-11e8-8dde-6926050f3fe8.png)

- [x] 2) It says copy 0 files!  It should copy any selected files, or if none are selected, then copy the # of files in the query.  It looks like this may be a result of the data shape change - its looking for UUID but it should be looking for ```latest_did``` 
<img width="1065" alt="screen shot 2018-05-28 at 5 48 32 pm" src="https://user-images.githubusercontent.com/15824252/40630284-6abfdc80-629f-11e8-808e-0e83e9d0b77d.png">
<img width="1036" alt="screen shot 2018-05-28 at 6 05 44 pm" src="https://user-images.githubusercontent.com/15824252/40630591-cca52a48-62a1-11e8-85e6-7e961719bfa8.png">

rosibaj:
@hlminh2000 Thanks for fixing this!

PR #454, opened 2018-06-05
🐛 Fix/update download id field

hlminh2000:


src/uikit/Input.js
```diff
@@ -2,4 +2,6 @@ import styled from 'react-emotion';
 
 export default styled('input')`
   ${({ theme }) => theme.input};
+  ${({ italic }) => (italic ? `font-style: italic` : ``)};
+  ${({ flex }) => (flex ? `flex: ${flex}` : '')};
```
jgnieuwhof: should we just introduce styled-system? https://github.com/jxnblk/styled-system/blob/master/docs/table.md

actually, couldn't find font-style, but it seems to do almost everything else

src/components/cavatica/CavaticaFileSummary.js
```diff
@@ -51,7 +51,7 @@ const enhance = compose(
           if (!studies[study]) {
             studies[study] = [];
           }
-          studies[study].push(file.node.uuid);
+          studies[study].push(file.node.kf_id);
```
jgnieuwhof: this is `kf_id` rather than `latest_did` intentionally?

hlminh2000: That's what's quoted in issue report, but @rosibaj could you confirm just in case ? :) https://github.com/kids-first/kf-portal-ui/issues/430

rosibaj: ```latest_did``` is correct - sorry about that @hlminh2000 - i correct #434 but not #430 when i sent it over to you today

hlminh2000: ah thanks @rosibaj!!! will update!

src/uikit/Input.js
hlminh2000: yeah I'll do that

PR #455, opened 2018-06-06
Updates (modularizes) theme typographic system

bdolly:
**MOTIVATION**
Create a more consistent, maintainable, and global typographic system for Kids First branding in order to make the branding/style easily applicable to assets outside the scope of the portal (website, styleguide, print, design assets,  Kids First sub-apps, etc) Replace narrowly scoped styled components with a system of composable global styles. This allows for more consistent font weights and font sizes across pages/views which can be update in one place and reflected across the entire project. 

**IMPLEMENTATION: Proposal 1**
Move typographic elements in `theme/defaultTheme` to their own file. Replace the class map style of using emotion's `css` template tag in favor of using `withTheme` style composition in components.

````js 
//theme/typography.js
export const headings = {
   ...otherHeadings, 
  h2: `
    ${headingsBase}
    color: ${colors.secondary};
    font-size: 30px;
    line-height: 0.95;
    letter-spacing: 0.4px;
    color: #2b388f;
    font-weight: 500;
  `,
}

export const text = {
  center: 'text-align: center;',
  left: 'text-align: left;',
  right: 'text-align: right;',
  ...otherTextUtils
};
````

````js 
// compoments/myArticle.js

const MyArticle = ({theme}) => (
   <article >
       <h2 css={`${theme.h2} `} >My Article Title</h2>
      <p css={`${theme.paragraph} ${theme.text.center}`}> article copy goes here .....  </p>
   </article>
);

export default withTheme(MyArticle);
````

**IMPLEMENTATION: Proposal 2**
Create a set of standard typographic components in `uiKit/Typography.js` that use the theme styles. 
````js 
// uiKit/Typography.js

export const H2 = styled.h1`
 ${({theme}) => theme.h2}
`
export const P = styled.p`
  ${({theme}) => theme.paragraph}
`;
 ...otherStyledComponents
````

````js 
// compoments/myArticle.js
import {H2, P} from '.../uiKit/Typography';
const MyArticle = ({theme}) => (
   <article >
       <H2 >My Article Title</H2>
      <P css={`${theme.text.center}`}> article copy goes here .....  </P>
   </article>
);

export default withTheme(MyArticle);
````
This second proposal could get a bit annoying having to constantly import standard elements but could have the advantage of standardizing typography components and have a spec in the styleguide and storybook for those elements.

### CODE REVIEW

File: src/uikit/EditableLabel.js [hunk 1]
```diff
@@ -173,15 +173,17 @@ export default compose(
             cursor: 'text',
           }}
         >
-          <span
+          <p
```
jgnieuwhof: As mentioned below could use `P` component from uikit rather than using `theme.paragraph`

src/uikit/Typography.js
```diff
@@ -0,0 +1,33 @@
+import styled from 'react-emotion';
+
+
+const HeadingBaseStyles = `
+  
+`;
+
+export const HeadingH2 = styled('h2')`
```
jgnieuwhof: I really like this idea of having componentized theme elements, something like the following:
```
export const P = styled('p')`...`;

export const H1 = styled('h1')`...`;

...
```

if we do that for all elements, and instead of importing the theme into CSS in the JSX we can just import these components and use them throughout the app

src/theme/typography.js
```diff
@@ -0,0 +1,94 @@
...
+
+export const text = {
+  center: 'text-align: center;',
+  left: 'text-align: left;',
+  right: 'text-align: right;',
+  underline: 'text-decoration: underline;',
+  italic: 'font-style: italic;',
+  captialize: 'text-transform: captialize;',
+  small: 'font-size: 75%;',
+};
```
jgnieuwhof: this is starting to look like styled-system territory, just don't want to re-invent the wheel here. https://github.com/jxnblk/styled-system/blob/master/docs/table.md

src/components/Join.js
```diff
@@ -61,7 +61,7 @@ const Consent = compose(
   }) => {
     return (
       <ConsentContainer>
-        <h3 className={theme.h3}>
+        <h3 css={`${theme.h3}`}>
```
jgnieuwhof: I've got a comment near the bottom discussing this pattern, I think we should just add proper `P`, `Hx` components to the uikit so we don't need to go through `css` / `theme` every time

src/components/Join.js
```diff
@@ -169,26 +169,26 @@ const JoinContent = compose(
 )(({ state: { loggedInUser }, effects: { setToast, closeToast }, history, theme, api }) => (
   <JoinContainer>
     <div className={`${theme.card} ${theme.column} `}>
-      <h2 className={theme.h2}>
+      <h2 css={`${theme.h2} ${theme.text.center}`}>
```
jgnieuwhof: combination of `H2` component & `styled-system` would really tidy this up

src/components/Join.js
```diff
@@ -208,15 +208,15 @@ const JoinContent = compose(
             title: 'Basic Info',
             render: ({ disableNextStep, nextStep, prevStep, nextDisabled, prevDisabled }) => (
               <div className={theme.column}>
-                <h3 className={theme.h3}>
+                <h3 css={`${theme.h3} margin-bottom:0;`}>
                   <Trans i18nKey="join.wizard.basicInfoHeader">Tell us about yourself</Trans>
                 </h3>
-                <p>
+                <h5 css={`${theme.h5} margin-top: 5px;`}>
```
jgnieuwhof: same here, styled-system + uikit would look like: `<H5 mt={5}>...</H5>`

jgnieuwhof:
@bdolly - LOL I actually missed your second proposal in the PR write up. I think Proposal 2 is the way to go. Proposal 2 + jamming the `styled-system` helpers in the typography uikit components should give us everything we need in a super clean way :)

hlminh2000:
So this PR seems to have diverged quite a bit from master, but as @jgnieuwhof mentioned the second approach would be well in-line with what we have going on :) with `styled-systems`, a lot of the "ad-hoc" stuff can be inlined too.

Also an additional advantage that can be implemented, `styled-components` allows us to do `<Button action/>` and `<Button hollow/>` so we can import the same component rather than having to import separate components too

jgnieuwhof:
for reference, https://github.com/jxnblk/styled-system/blob/master/docs/api.md#complex-styles

bdolly:
Closing in favor of styling api updates in #497

PR #458, opened 2018-06-06
✨ Hot Module Reloading

bdolly:
Attempting to setup react-hot-loader so there is no page refresh on save.


bdolly:
@jgnieuwhof got this working by just turning off Webpack dev server page refresh in the `config-overrides`.

src/components/SplashPage.js
```diff
@@ -8,6 +8,7 @@ const SplashPage = compose(withTheme)(({ theme, children }) => (
     css={`
       width: 630px;
       margin: auto;
+      border: 3px solid green;
```
hlminh2000: 🔥

bdolly:
@hlminh2000 I think this is g2g, not sure what's up with the netlify status blocking it thought

hlminh2000:
@bdolly it was just missing some dependencies from `package.json` :P ran a test local prod build and it works after some installs. Netlify is working now too but not sure why it's not reporting back to github... Anyway, merging now :)

PR #459, opened 2018-06-07
✨ adds connected with button to user profile page

jgnieuwhof:
Did a bunch more css clean up, and added a 'connected with' button to the profile/settings page.

Since we don't lock a user to a specific method of logging in it simply uses the last method of login (storing the last method in localStorage to account for the re-load w/ valid token case).

### DIFF
--- a/src/components/UserProfile/AboutMe.js
+++ b/src/components/UserProfile/AboutMe.js
@@ -267,7 +267,7 @@ export default compose(
                   }}
                 />
               ) : (
-                <React.Fragment>
+                <Flex>
                   <HollowButton
                     onClick={() => {
                       setWebsite(profile.website || '');
@@ -279,9 +279,6 @@ export default compose(
                     Cancel
                   </HollowButton>
                   <SaveButton
-                    css={`
-                      height: 30px;
-                    `}
                     onClick={async () => {
                       await submit({
                         website,
@@ -296,7 +293,7 @@ export default compose(
                   >
                     Save
                   </SaveButton>
-                </React.Fragment>
+                </Flex>
               ))}
           </H2>
           <div>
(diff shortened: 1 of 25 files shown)


src/components/UserProfile/AboutMe.js
```diff
@@ -1,45 +1,68 @@
...
-const ClickToAdd = ({ theme, children, ...rest }) => (
-  <a className={`clickToAdd`} {...rest}>
-    {children}
-  </a>
-);
+`;
+
+const AboutMeContainer = styled(Container)`
+  ${({ theme }) => theme.row};
+`;
```
hlminh2000: Maybe we can add `${({row}) => row ? theme.row : ''}` so we can do `<Container row>` in component? seems like a pretty re-usable pattern.

src/components/UserProfile/Settings.js
```diff
@@ -1,46 +1,27 @@
...
-          <ul css={theme.verticalNav}>
-            <li>
-              <a className={mode === 'account' ? 'active' : ''} onClick={() => setMode('account')}>
-                <Trans>Settings</Trans>
-              </a>
-            </li>
-          </ul>
-        </div>
+    <Flex justifyContent="center" pt={4} pb={4}>
+      <Container flex flexDirection="row" alignItems="flex-start">
```
hlminh2000: `<Container row alignItems="flex-start"/>` :)

src/components/UserProfile/index.js
```diff
@@ -141,7 +105,7 @@ export default compose(
         justify-content: center;
       `}
     >
-      <Container className={theme.row}>
+      <Container className={theme.row} alignItems="center">
```
hlminh2000: <Container row alignItems="center"/>

PR #450, opened 2018-06-01
💄 Overall file repo styling touchups

hlminh2000:
was going to work on all of https://github.com/kids-first/kf-portal-ui/issues/63 but figured to take it one step at a time so the PRs are more managable.

Actually this pretty much covers everything in that issue now...

PR #461, opened 2018-06-07
🐛 fix broken portal

hlminh2000:
This reverts commit 00ee27686f205c08f2feab84043dd700aab5d15d (which reverts #450 in a hotfix on master).

and tries to fix whatever that's broken... not sure what yet


hlminh2000:
yeah totally :) will be using styled system a lot more going forward ;)

jgnieuwhof:
Same!

PR #479, opened 2018-06-15
✨ add a beta message to the join wizard

jgnieuwhof:


hlminh2000:
👍 Just had to fix some sizing that results in smaller screens :) Otherwise, clean! Thank you!!!

PR #527, opened 2018-07-11
modal positioning for UserSnap

bdolly:
updating modal positioning so UserSnap accurately captures screenshots

### DIFF
--- a/src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
+++ b/src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
@@ -55,7 +55,10 @@ const IdFilterContainer = styled(Column)`
   }
 `;
 
-const AggregationSidebar = compose(injectState, withTheme)(
+const AggregationSidebar = compose(
+  injectState,
+  withTheme,
+)(
   ({
     state,
     effects,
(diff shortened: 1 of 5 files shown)


bdolly:
@hlminh2000 this should be good to merge, suppper minor tweak to the dropdown styles

src/uikit/Dropdown/ui.js
```diff
@@ -34,18 +34,19 @@ export const DropdownOptionsContainer = styled(Column)`
   min-width: 100%;
   z-index: 1;
   cursor: pointer;
-  padding: 5,
-  right: ${({ align }) => (align === 'right' ? 0 : 'auto')};
-  left: ${({ align }) => (align === 'right' ? 'auto' : 0)};
```
hlminh2000: 👍

src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
```diff
@@ -91,34 +94,44 @@ const AggregationSidebar = compose(injectState, withTheme)(
                         ...props,
                         translateSQONValue,
                         closeModal: effects.unsetModal,
+                        onFacetNavigation: path => {
+                          let facetNavEvent = {
+                            category: TRACKING_EVENTS.categories.fileRepo.filters + ' - Advanced',
+                            action: TRACKING_EVENTS.actions.click + ' side navigation',
+                            label: path,
+                          };
+                          trackFileRepoInteraction(facetNavEvent);
```
hlminh2000: doesn't seem it's breaking anything here but for safety, it probably should be standard to check on all these callbacks coming from props to call them as well when it's spread down like here, otherwise this would override anything passed down from above

bdolly:
closing this as the issue is stale and the use of UserSnap will soon be deprecated in place of Hotjar

PR #554, opened 2018-07-18
🐛 fixes the fields

hlminh2000:


src/components/forms/BasicInfoForm.js
```diff
@@ -410,24 +410,33 @@ export default compose(
...
+                    ).long_name;
+                    const route = (
+                      results[0].address_components.find(c => c.types.includes('route')) ||
+                      defaultObject
+                    ).long_name;
+                    const zip = (
+                      results[0].address_components.find(
+                        c => c.types.includes('zip') || c.types.includes('postal_code'),
+                      ) || defaultObject
+                    ).long_name;
```
hlminh2000: seems a slightly different prettier setting? :/

cy: yeah let's just go with yours

hlminh2000: 👍

PR #578, opened 2018-07-26
🐛 removes `egoGroups` property from loggedInUser state

bdolly:


src/services/analyticsTracking.js
```diff
@@ -97,7 +97,7 @@ const setUserDimensions = (userId, role, groups) => {
   }
 
   if ((groups && groups.length) || (GAState.egoGroups && GAState.egoGroups.length)) {
-    let _groups = JSON.stringify({ egoGroups: groups || GAState.egoGroups });
+    let _groups = JSON.stringify({ egoGroups: groups[0] || GAState.egoGroups[0] });
```
hlminh2000: I'm getting this error here
screen shot 2018-07-26 at 4 57 02 pm

hlminh2000:
I was still getting that same error so pushed a fix for it, but now got another one during account creation...

screen shot 2018-07-26 at 5 41 54 pm

src/services/analyticsTracking.js
```diff
@@ -97,7 +97,7 @@ const setUserDimensions = (userId, role, groups) => {
   }
 
   if ((groups && groups.length) || (GAState.egoGroups && GAState.egoGroups.length)) {
-    let _groups = JSON.stringify({ egoGroups: groups[0] || GAState.egoGroups[0] });
+    let _groups = JSON.stringify({ egoGroups: groups ? groups[0] : GAState.egoGroups[0] });
```
hlminh2000: groups was undefined here. It's an OR condition so that can happen...

bdolly: What if just set the GAState.egoGroups default of `[" "]` that way it will pass through and we don't have to worry about doing so many conditional? 

I've been meaning to refactor this whole service with better patterns but haven't had time yet.

hlminh2000: It's actually `groups` that's undefined so `groups[0]` would throw an error. I'm not familiar with the rest of this so did this just to fix that one error without touching the rest and break things :P

PR #587, opened 2018-07-30
💄 Style/495 user settings and profile

ciaranschutte:


src/components/UserProfile/ConnectedWithBadge.js
```diff
@@ -20,22 +19,21 @@ const icons = {
   [FACEBOOK]: x => <FBIcon color="#428bca" {...x} />,
 };
 
-const ButtonItem = styled(x => <Box p={3} {...x} />)`
-  border-right: ${({ divider, theme }) => (divider ? `1px solid ${theme.greyScale8}` : `none`)};
+const Status = styled(x => <Span p={2} {...x} />)`
```
hlminh2000: maybe this for simplicity?
```
styled(Span)`
  padding: 2px;
  ...
`
```

src/components/UserProfile/UserIntegrations/index.js
```diff
@@ -69,6 +70,31 @@ const Gen3DetailButton = styled(ConnectedButton)`
...
+        margin-right: 5px;
+      `}
+    />
+    Connect
+    <RightIcon
+      size={14}
+      css={`
+        margin-left: 14px;
+      `}
+    />
```
hlminh2000: I think we prefer splitting these into styled components instead of inline styles now, or styled-system. There's a function called `applyDefaultStyles` you can import from `src/uikit/Core.js` which basically provides all the styled-system functions, so if you wrap the base components with that, you can just use some quick props to do the small styles ;)

src/components/UserProfile/ui/index.js
```diff
@@ -95,6 +95,10 @@ export const ClickToAdd = styled('a')`
   font-size: 12px;
   text-decoration: underline;
   color: ${({ theme }) => theme.primary};
+
+  &:hover {
+    cursor: pointer
+  }
```
hlminh2000: do we actually need this wrapped in `:hover`?

src/components/UserProfile/UserIntegrations/index.js
ciaranschutte: More like...
```
const ConnectButton = ({ ...props }) => {
  const ExternalLink = applyDefaultStyles(ExternalLinkIcon);
  const RightArrow = applyDefaultStyles(RightIcon);

  return (
    <ActionButton {...props} px={6} br={19}>
      <ExternalLink size={12} ml={10} mr={5} />
      Connect
      <RightArrow size={14} ml={14} />
    </ActionButton>
  );
}
```

hlminh2000: yeah, or just wrap `applyDefaultStyles` around the `ExternalLinkIcon` and `RightIcon` themselves? personally I think this looks cleaner :P

PR #590, opened 2018-07-31
💄 Move Footer into content wrapper so it's not fixed

ciaranschutte:


hlminh2000:
I think for this we want to maintain the existing behaviour for the File Repo page actually :P

PR #605, opened 2018-08-02
Style/512 fix footer size

ciaranschutte:
Use 100vh for fixed footer pages Dashboard and FileRepo


src/components/Page.js
```diff
@@ -3,14 +3,14 @@ import Header from 'components/Header';
 import Footer from 'components/Footer';
 import styled from 'react-emotion';
 import Column from 'uikit/Column';
+import { applyDefaultStyles } from '../uikit/Core';
```
hlminh2000: oh you can import it as `uikit/Core` btw :)

PR #607, opened 2018-08-02
Style/495 user settings and profile

ciaranschutte:


ciaranschutte:
Formatting changes occuring in this so you might want to hold off until I fix my save on formatting

PR #610, opened 2018-08-03
⚡️optimizes fetching data for the download columns and some styling

hlminh2000:
- make the loader smaller
- only fetch when cell value changes


hlminh2000:
@rosibaj the netlify listed by Github turned out to be the prod one, so here's the QA one that would do the trick: https://5b6452323813f03830b8f7ba--kf-qa-portal.netlify.com/

hlminh2000:
alternatively you can grab your token from the prod portal and use the link listed by github, that would let you test against prod

rosibaj:
@hlminh2000 This looks great!  Table bump is fixed on selection and filter reload.  Lets merge!!

PR #615, opened 2018-08-03
:lipstick: :arrow_up: User menu dropdown

ciaranschutte:


ciaranschutte:
Bumping version helped track down the issue, only minor update so shouldn't be an issue

PR #627, opened 2018-08-08
💄 Style/582 update profile icons

ciaranschutte:
- Updated icons and labels
- Added new Patient/Family role


ciaranschutte:
@hlminh2000 good catch :)

rosibaj:
@hlminh2000 we do not have defined fields right now - lets use the same set for now.

PR #655, opened 2018-08-17
💄 new profile fields

ciaranschutte:


src/common/constants.js
```diff
@@ -101,7 +101,7 @@ export const DISEASE_AREAS = [
   'Osteosarcoma',
   'Structural Heart & Other Defects',
   'Syndromic Cranial Dysinnervation Disorders',
-  'CBTTC/PNOC Brain Tumor Dataset',
+  'Pediatric Brain Tumors',
```
hlminh2000: I personally prefer `text-transform: capitalize;` for this stuff, what do you think?

src/components/UserProfile/FindMe.js
```diff
@@ -143,7 +151,7 @@ const socialItems = {
         `}
       />
     ),
-    name: 'ORCHID ID',
+    name: 'ORCID iD:',
```
hlminh2000: damn that's trippy :P

src/common/constants.js
ciaranschutte: An issue with `text-transform: capitalize` is that we end up with `Select An Option` instead of `Select an option` for a default.

Or, more concretely, if say the third item needs to be `Patients with both childhood....`instead of `Patients With Both Childhood`. We have this central file to specify vs a bunch of CSS rules to target specific children across potentially multiple select forms.

PR #663, opened 2018-08-21
📈 Gen3 tracking update

bdolly:
adds a label to GA so we can track Gen3 connections failures and success events


hlminh2000:
already merged it seems, so closing. but how did this get merged with the PR still open? @bdolly

PR #665, opened 2018-08-21
📈 adds immediate dom content to user snap

hlminh2000:
Putting on my mad scientist hat to do an extreme log measure :P

Can't reproduce the google button issue, so adding this for when the issue is reported again so we can see what is and isn't actually in the dom

### DIFF
--- a/src/services/usersnap.js
+++ b/src/services/usersnap.js
@@ -2,6 +2,7 @@ import scriptjs from 'scriptjs';
 import urlJoin from 'url-join';
 import { merge, pick } from 'lodash';
 import { UI_VERSION } from 'common/constants';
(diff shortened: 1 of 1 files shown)


hlminh2000:
@bdolly would be great to get your thoughts on this. I think it'll be useful in situations like the google button issue, but maybe it's too extreme? lmk

bdolly:
@hlminh2000 so this just adds the initial rendered DOM as a string to the usersnap meta?

hlminh2000:
It should be adding it just before sending the data no?

bdolly:
@hlminh2000 yeah, I think that's how it should work. I think it should be good to roll out and see how it does.

PR #670, opened 2018-08-23
💩temporary implementation of split tabbed aggs in the portal

hlminh2000:
We'll have to think about ways to support this from arranger later, putting this here for now.


hlminh2000:
QA preview: https://5b7f1fc702ed831e40fc73bc--kf-qa-portal.netlify.com/

PR #683, opened 2018-08-27
📈 tracking events for  query abandonment

bdolly:
simple events to track when people bail out of completing query actions

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -20,6 +20,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
+
     <title>Portal</title>
   </head>
   <body>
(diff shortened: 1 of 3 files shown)


public/index.html
```diff
@@ -38,5 +39,13 @@
...
     -->
     <script src="https://apis.google.com/js/platform.js"></script>
+    <script>
+      <!-- track when users exit the site after starting a query -->
+      window.onbeforeunload = function () {
+        if(window.localStorage.getItem('KF_GA_TIMING_INIT_FILE_QUERY_TO_CAVATICA_COPY') || window.localStorage.getItem('KF_GA_TIMING_INIT_FILE_QUERY_TO_DOWNLOAD') ){
+          window.ga('send', 'event', 'File Repo', 'Query Abandoned', 'exited site', 3);
+        }
+      }
+    </script>
```
hlminh2000: does this have to be in the html? why not have this in `src/services/analyticsTracking.js`?

bdolly: couldn't get it to fire from the analytics tracking service. I think it's because the event is fired after the page unloads all it's external resources before exiting

hlminh2000:
@bdolly I pushed a hotfix as it was breaking logins, that's causing a conflict for this PR so if you could take a look at that while rebasing this it'd be great :)

https://github.com/kids-first/kf-portal-ui/pull/702

bdolly:
@hlminh2000 good catch, I thought we had worked out the login issue. I rebased into this branch so should be g2g

PR #685, opened 2018-08-27
📈track the Gen3 authorized studies per user

bdolly:


src/services/analyticsTracking.js
```diff
@@ -85,6 +85,11 @@ export const initAnalyticsTracking = () => {
   });
 };
 
+export const setUserDimension = (dimension, val) => {
+  if (isObject(val)) val = JSON.stringify(val);
+  ReactGA.set({ [dimension]: val });
+};
```
hlminh2000: I don't think it breaks in this case, but I much prefer to write this as following to remove room for potential mutative code later:
```
export const setUserDimension = (dimension, val) => {
  const serialized = isObject(val) ? JSON.stringify(val) : val
  ReactGA.set({ [dimension]: serialized });
};
```

Issue #613, opened 2018-08-03
Header: Adding an "All apps" menu dropdown


kcullion:
The Kids First item in the header will open a dropdown with links to the following.  
1. The Data Resource Portal (this will be always active when on the portal)
2. Website - will link to: https://kidsfirstdrc.org/ (OPEN IN A NEW WINDOW)
3. Support - will link to: https://kidsfirstdrc.org/support/getting-started/ (OPEN IN A NEW WINDOW) 
4. Contact - will link to: https://kidsfirstdrc.org/contact

It looks like this:
![screen shot 2018-08-23 at 3 19 53 pm](https://user-images.githubusercontent.com/18424138/44547110-04536d80-a6e8-11e8-826d-582efa18a574.png)

Mockups can be found here:
Off state: https://zpl.io/VxYolPw
On state: https://zpl.io/V0yWqKK

Icons can be downloaded here:
[KF-all-apps-icons.zip](https://github.com/kids-first/kf-portal-ui/files/2291179/KF-all-apps-icons.zip)

ciaranschutte:
@rosibaj https://5b86ea64c96592716fc9dfe2--kf-qa-portal.netlify.com/

rosibaj:
@ciaranschutte i think the colors are off.  

- [ ] Can you change all headings to #90278e 
EX:

![screen shot 2018-08-29 at 3 50 22 pm](https://user-images.githubusercontent.com/15824252/44811792-85f14280-aba3-11e8-93d4-59d9a15afab8.png)

![screen shot 2018-08-29 at 3 51 08 pm](https://user-images.githubusercontent.com/15824252/44811769-7bcf4400-aba3-11e8-9503-aa6adb491486.png)

- [ ] When they user mouses over the menu or user profile it should be icon` v` and words and icon highlighed to #c03299

- [ ] When they user clicks the menu or user profile it should be icon `^` and words and icons highlight to #c03299

ciaranschutte:
https://5b87d9483813f076d0b702e1--kf-portal.netlify.com/
@rosibaj should be good now :)

ciaranschutte:
@rosibaj https://5b87d9483813f076d0b702e1--kf-portal.netlify.com/

rosibaj:
@ciaranschutte When i login on this link I see this now:
[...]

PR #689, opened 2018-08-29
✨ #613 all apps menu dropdown

ciaranschutte:


src/components/Header/AppsMenu.js
```diff
@@ -0,0 +1,78 @@
...
+  DropdownExternalLink,
+  DropdownRow,
+  AllAppsLabelContainer,
+  MenuLabelContainer,
+} from './ui';
+import AllAppsContactIcon from '../../icons/AllAppsContactIcon';
+import AllAppsSupportIcon from '../../icons/AllAppsSupportIcon';
+import AllAppsWebsiteIcon from '../../icons/AllAppsWebsiteIcon';
+import AllAppsPortalIcon from '../../icons/AllAppsPortalIcon';
+import AllAppsMenuIcon from '../../icons/AllAppsMenuIcon';
```
hlminh2000: can probably be referenced as `icons/...`

src/components/Header/AppsMenu.js
```diff
@@ -0,0 +1,78 @@
...
+);
+
+export default compose(
+  withState('isDropdownVisible', 'setDropdownVisibility', false),
+  withHandlers({
+    toggleDropdown: ({ isDropdownVisible, setDropdownVisibility }) => e => {
+      setDropdownVisibility(!isDropdownVisible);
+    },
+  }),
+)(AppsMenu);
```
hlminh2000: I prefer render props over hoc personally, I find it actually reduces the cost of moving state around much easier... In this case, I'd just use `react-component-component` for these kinds of state

Also for your entertainment, check out this article Chang wrote a while back about the namespacing problem with hoc :
https://hackernoon.com/solving-the-problems-of-higher-order-components-without-throwing-the-baby-out-with-the-bathwater-40ddc72df5aa

src/components/Header/ui.js
```diff
@@ -4,9 +4,12 @@ import Row from 'uikit/Row';
 import Gravtar from 'uikit/Gravatar';
 import { DropdownContainer, DropdownOptionsContainer } from 'uikit/Dropdown';
 import { space } from 'styled-system';
+import { DropdownLabelContainer } from '../../uikit/Dropdown/ui';
```
hlminh2000: `uikit/Dropdown/ui`? :)

src/components/Header/AppsMenu.js
ciaranschutte: ```const hocToProvider = hoc => hoc(({ render, ...props }) => render(props));```

Issue #696, opened 2018-08-30
Share button does not trigger dropdown in Firefox.


andricDu:
- got to firefox
- Make a query in the dashboard
- Click share -- nothing happens

andricDu:
#718

PR #718, opened 2018-08-31
Uses component state to manage tooltip.

andricDu:
This fixes the issue of the tooltip not showing up in firefox.


andricDu:
#696

Issue #606, opened 2018-08-02
Overall style consistency: headings + copy


kcullion:
Mission: to make the headings + copy consistent throughout the portal.
(Note: some font size variation occurs in the File Repository in order to keep the interface clean)

Fixes outlined in screenshots, please download:
[portal-headings-copy.zip](https://github.com/kids-first/kf-portal-ui/files/2254956/portal-headings-copy.zip)

Issue #599, opened 2018-08-01
Overall style consistency: Teal Buttons


kcullion:
There should be only 1 style of teal button throughout the portal (with slight variation - small and large). Please fix all teal buttons so they have the following style:

Style:
![teal-buttons](https://user-images.githubusercontent.com/18424138/43539602-56aca620-9593-11e8-934d-2910ab9be7ef.png)

### **Instances of small teal buttons**
- [x] Profile - Save buttons
![screen shot 2018-08-01 at 1 54 45 pm](https://user-images.githubusercontent.com/18424138/43539278-7754e226-9592-11e8-8a36-185456e687e3.png)

- [x] Modal save buttons 
Profile edit basic info: 
![screen shot 2018-08-01 at 1 53 57 pm](https://user-images.githubusercontent.com/18424138/43539300-823b3bf4-9592-11e8-8e7e-18f5abb65455.png)

- [x] Profile - connect to Cavatica modal
![screen shot 2018-08-01 at 2 13 11 pm](https://user-images.githubusercontent.com/18424138/43540193-092c025e-9595-11e8-8ffa-dd67db60da12.png)

- [x] File Repository - all filters button
![screen shot 2018-08-01 at 2 01 58 pm](https://user-images.githubusercontent.com/18424138/43539692-8db78f0e-9593-11e8-87ae-88a18c25646c.png)

And view results button (in all filters modal):
![screen shot 2018-08-01 at 2 02 11 pm](https://user-images.githubusercontent.com/18424138/43539719-a5b2c948-9593-11e8-8bf8-5b51e402a5c4.png)

- [x] File Repository - Download button (also has download icon)
![screen shot 2018-08-01 at 2 02 03 pm](https://user-images.githubusercontent.com/18424138/43539704-9897c0d8-9593-11e8-96d8-56dc69d7ae12.png)

And download manifest modal button:
![screen shot 2018-08-01 at 2 03 48 pm](https://user-images.githubusercontent.com/18424138/43539750-c3f12648-9593-11e8-93a6-82e5a78a87ea.png)

- [x] File Repository - Copy files to cavatica modal:
![screen shot 2018-08-01 at 2 04 52 pm](https://user-images.githubusercontent.com/18424138/43539803-ecd6ea34-9593-11e8-9796-fd64a77b8195.png)

- [ ] File Repository - Upload IDs modal:
[...]

Issue #596, opened 2018-08-01
Overall style consistency: White Buttons


kcullion:
There should be only 1 style of white button throughout the portal. Please fix all white buttons so they have the following style

**White Action buttons:**
Style:
![screen shot 2018-08-01 at 1 30 24 pm](https://user-images.githubusercontent.com/18424138/43538138-27e307ac-958f-11e8-8340-9caff26c985b.png)

### **Instances of white buttons**
- [x] Dashboard - Profile buttons
![screen shot 2018-08-01 at 11 21 43 am](https://user-images.githubusercontent.com/18424138/43531149-2e8d088a-957d-11e8-806f-58d3929cf184.png)
- [x] Profile edit/cancel

![screen shot 2018-08-01 at 11 56 53 am](https://user-images.githubusercontent.com/18424138/43533219-4b80033e-9582-11e8-9265-48ba85e3314f.png)

![screen shot 2018-08-01 at 11 58 29 am](https://user-images.githubusercontent.com/18424138/43533188-3ebb2890-9582-11e8-8c8e-8025f0bcc874.png)

- [x] User Settings page:
![screen shot 2018-08-01 at 12 12 55 pm](https://user-images.githubusercontent.com/18424138/43533993-3ffcfdb2-9584-11e8-9d03-80b104d9bf70.png)

- [x] ALL Modal Cancel buttons (I know this isn't the white button in my mockup, but since we are using white cancel buttons in the profile widgets, please change this to the white button style)

![screen shot 2018-08-01 at 12 00 22 pm](https://user-images.githubusercontent.com/18424138/43533492-e9a5357a-9582-11e8-8abd-d10c55ddcf41.png)
![screen shot 2018-08-01 at 12 19 57 pm](https://user-images.githubusercontent.com/18424138/43534402-415cde42-9585-11e8-84ec-b839b3b245dc.png)
(All filters cancel) 
![screen shot 2018-08-01 at 1 28 43 pm](https://user-images.githubusercontent.com/18424138/43538019-e616b72e-958e-11e8-9a95-aa7f51a75902.png)

- [x] Change avatar button - also try to make the label fit on one line  (may mean wider modal, which is fine)
![screen shot 2018-08-01 at 12 13 54 pm](https://user-images.githubusercontent.com/18424138/43534100-801c6fea-9584-11e8-88fe-fee3f74a4647.png)

- [x] File Repository
[...]

PR #721, opened 2018-09-03
Style/design updates

ciaranschutte:


src/components/ShareSaveQuery/ui.js
```diff
@@ -11,3 +12,11 @@ export const CustomLightButton = styled(LightButton)`
   min-width: 75px;
   font-size: 10px;
```
hlminh2000: For deprecated components in the future, let's add deprecation messages, maybe like so:
```
export const CustomLightButton = styled(LightButton)`
  ${() => console.warn("CustomLightButton has been deprecated (src/components/FileRepoSidebar/ui.js)")}
  min-width: 75px;	  min-width: 75px;
  font-size: 10px;	  font-size: 10px;
```

src/components/UploadIdsModal.js
```diff
@@ -2,16 +2,17 @@ import React from 'react';
...
 import { ModalFooter } from './Modal';
+import { TealActionButton } from 'uikit/Button';
+import { Paragraph } from 'uikit/Core';
+import { FileRepoH3 as H3 } from 'uikit/Headings';
+import { TableHeader } from 'uikit/Table';
 
-const UploadButton = withTheme(({ theme, ...props }) => (
-  <button className={theme.actionButton} {...props} />
-));
+const UploadButton = withTheme(({ theme, ...props }) => <TealActionButton {...props} />);
```
hlminh2000: this component can go, it's not exported so no risk of breaking stuff. `TealActionButton` can be consumed directly now

src/components/UserDashboard/index.js
```diff
@@ -55,21 +56,25 @@ export default compose(
...
       >
-        <H2 pb={10}>Welcome, {loggedInUser.firstName}!</H2>
+        <H1 mb={'35px' /* On removing Beta Message change to 44px */}>
+          Welcome, {loggedInUser.firstName}!
+        </H1>
         <Column>
           <PromptMessageContainer info my={20}>
             <PromptMessageHeading info mb={10} color={theme.secondary}>
-              The Portal is currently in <strong>BETA Phase</strong>
+              The Portal is currently in <H2 display="inline-block">BETA Phase</H2>
```
hlminh2000: just a nit-pick, not sure I entirely agree with the semantic of having a header inside a header, isn't the whole thing a header and this part is just bolded? #problemForLater :P

src/components/ShareSaveQuery/ui.js
hlminh2000: of course with a little more engineering so we don't have these in prod :P

hlminh2000:
This PR addresses a lot of things in #599, #606 and #596. @rosibaj @kcullion we can probably take a review of the left-over stuff from those tickets and move to a new one.

This PR was getting a little too big so we wrapped it up before it explodes further, but let's try to keep it to 1 ticket per PR for future :)

kcullion:
Looks great @ciaranschutte! As I mentioned, it's just the "settings" (dashboard) and "connect" (settings) buttons that look quite large.

PR #722, opened 2018-09-03
💄 Contact info fixes

ciaranschutte:


hlminh2000:
LGTM :) for future reference though, we can give these guys a shot if we run into more problems with phone numbers: https://github.com/catamphetamine/libphonenumber-js

ciaranschutte:
@hlminh2000 yup checked out that lib  ..  looks good ... we'd need more ui to specify country etc

PR #724, opened 2018-09-04
🐛 Bug/No modal overflow on FF

ciaranschutte:


src/components/forms/BasicInfoForm.js
```diff
@@ -206,14 +206,7 @@ export default compose(
...
     <Fragment>
-      <div
-        css={`
-          ${theme.row};
-          margin-bottom: 20px;
-          z-index: 1;
-          position: relative;
-        `}
-      >
+      <Row mb="20px" zIndex="1" position="relative" minHeight="0">
```
hlminh2000: I suggest we should consider enforcing styled components on anything that has more than 3 inline styles :P

PR #772, opened 2018-09-11
🔥 removes uploads ids filter dropdown

### LINKED ISSUE #758: Hide Upload id dropdown
This feature was developed with the expectation that manifests and saved manifest would be implemented as features in the portal. 

While this was implemented, it is not desirable for users to use the download client so we would like to hide the manifest portion of the feature (just the v dropdown) since users do not have a way to manage this data. 

- remove teh dropdown
- leave the rest of the feature in place, with the same functionality

bdolly:
Fixes #758 
updates wording on action buttons in uploads ids modal

### DIFF
--- a/src/components/UploadIdsModal.js
+++ b/src/components/UploadIdsModal.js
@@ -41,7 +41,7 @@ const UploadIdsModal = ({
       matchedEntity: <TableHeader>Matched Entity</TableHeader>,
       entityId: <TableHeader>Entity Id</TableHeader>,
     }}
-    browseButtonText={<Trans>Browse</Trans>}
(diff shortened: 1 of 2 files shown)


src/components/UploadIdsModal.js
```diff
@@ -61,7 +61,7 @@ const UploadIdsModal = ({
             setSQON(nextSQON);
             unsetModal();
           },
-          submitText: 'Upload',
```
hlminh2000: @rosibaj was this part of the ticket?

rosibaj: thanks @hlminh2000 looks fine lets merge

PR #797, opened 2018-09-24
Style/776 add social login error alerts

ciaranschutte:


hlminh2000:
Looks like on Chrome, if cookies is disabled we won't even get localStorage? If that's the case, we might need a screen like github below to inform users
screen shot 2018-09-26 at 5 24 40 pm

screen shot 2018-09-26 at 5 39 03 pm

ciaranschutte:
> Looks like on Chrome, if cookies is disabled we won't even get localStorage? If that's the case, we might need a screen like github below to inform users
> screen shot 2018-09-26 at 5 24 40 pm
> 
> screen shot 2018-09-26 at 5 39 03 pm

Great. I'll make a separate ticket for that because will need to chat with @kcullion for design

src/components/loginButtons/DisabledGoogleLogin.js
```diff
@@ -0,0 +1,58 @@
...
+`;
+
+const DisabledGoogleLogin = () => (
+  <DisabledLogin height="40px" width="240px">
+    <IconWrapper>
+      <GoogleIcon src={GIcon} />
+    </IconWrapper>
+    <ButtonText>Sign in with Google</ButtonText>
+  </DisabledLogin>
+);
```
hlminh2000: https://www.npmjs.com/package/react-social-login-buttons =))))

src/index.js
```diff
@@ -12,9 +12,9 @@ import MaintenancePage from './MaintenancePage';
 
 import './i18n';
 
-initAnalyticsTracking();
 googleSDK();
 facebookSDK();
+initAnalyticsTracking();
```
hlminh2000: just curious, why the move? :)

src/services/facebookSDK.js
```diff
@@ -1,18 +1,22 @@
...
+
+    global.FB.init({
+      appId: facebookAppId,
+      autoLogAppEvents: true,
+      xfbml: true,
+      version: 'v2.10',
+    });
+
+    resolve(global.FB);
+  });
```
hlminh2000: what was the problem with the original `scriptjs` approach?

src/components/loginButtons/DisabledFacebookLogin.js
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const DisabledGoogleLogin = () => (
+  <DisabledLogin height="40px" width="236px">
+    <IconWrapper>
+      <FBIcon />
+    </IconWrapper>
+    <ButtonText>Log in With Facebook</ButtonText>
+  </DisabledLogin>
+);
```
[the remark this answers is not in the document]
ciaranschutte: -__- after we even joked about a react package for everything!

src/services/facebookSDK.js
ciaranschutte: No error callback using the script loader

hlminh2000: I see, might make more sense to have this in `<script>` anyway for now because it's explicitly global, everywhere else, it's referring to `global.FB` :)

Ideally I think we should make this a lazy factory so components can access the FB instance through this rather than reading from `global.FB` but that can happen another time :P

PR #806, opened 2018-09-28
Bug/800 user tag updates

ciaranschutte:


ciaranschutte:
@hlminh2000 few additional fixes here when you get a chance please :)

package.json
```diff
@@ -25,12 +25,12 @@
     "md5": "^2.2.1",
     "popback": "^0.0.3",
     "query-string": "^5.1.0",
-    "react": "^16.3.0",
+    "react": "^16.5.2",
     "react-app-rewire-hot-loader": "^1.0.1",
     "react-autocomplete": "^1.8.1",
     "react-component-component": "^1.2.1",
     "react-copy-to-clipboard": "^5.0.1",
-    "react-dom": "^16.3.0",
```
hlminh2000: can we close the other PR then?

src/common/displayFormatters.js
```diff
@@ -23,3 +23,15 @@ export const formatPhoneNumber = phone => {
...
+ * Title cases string
+ * @param {String} string
+ * @returns {String}
+ */
+export const titleCase = (string = '') => {
+  return string.replace(
+    /\w\S*/g,
+    text => text.charAt(0).toUpperCase() + text.substr(1).toLowerCase(),
+  );
+};
```
hlminh2000: lodash's `startCase`? :)))

package.json
ciaranschutte: merged other pr into next

src/common/displayFormatters.js
ciaranschutte: let's leave this one as is. 
startCase would incorrectly remove hyphens from some terms eg. 'anti-inflammatory'

PR #829, opened 2018-10-05
Update styling

ciaranschutte:


src/components/Footer.js
```diff
@@ -38,10 +39,7 @@ const FooterContainer = styled('footer')`
   position: relative;
 `;
 
-const Details = styled('div')`
-  ${({ theme }) => theme.column};
-  ${({ theme }) => theme.center};
-`;
+const Details = styled(Row)``;
```
hlminh2000: can just use `Row` and remove this component then :P

PR #833, opened 2018-10-08
📝 Replace cra readme

dankolbman:
Replaces the CRA readme.


README.md
```diff
@@ -1,2173 +1,27 @@
...
-  const welcome = <h2>Welcome to React</h2>;
-  // expect(wrapper.contains(welcome)).to.equal(true);
-  expect(wrapper.contains(welcome)).toEqual(true);
-});
-```
-
-All Jest matchers are [extensively documented here](http://facebook.github.io/jest/docs/expect.html).<br>
-Nevertheless you can use a third-party assertion library like [Chai](http://chaijs.com/) if you want to, as described below.
-
-Additionally, you might find [jest-enzyme](https://github.com/blainekasten/enzyme-ma
```
hlminh2000: Just checking, are we sure this Acknowledgment section is no longer needed? :)

PR #859, opened 2018-10-17
📈 hotjar init implementation

bdolly:
adds the initial Hotjar tracking code with some minor refactoring of the analtyicsTracking service


bdolly:
@hlminh2000 that's a good point! As I was working through updating the end points I was thinking there had to be a better way to do that. I will work on refactoring that. I was also thinking of refactoring the whole service implementation using better patterns.

bdolly:
Closing in favor of #894 which addresses the issues identified above

PR #866, opened 2018-10-17
📈 updates abandoned query tracking

bdolly:
Track when a query abandonment when a user starts a query and navigates away from the page without saving or downloading it


bdolly:
🚧 block merge until #894 is merged

PR #894, opened 2018-10-18
🎨 Analytics tracking service reorg

bdolly:
re-organizes the structure of the analyticsTracking service to allow for easier integration with multiple analytics tracking providers.  

🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧
@hlminh2000 don't merge this to prod until we get the go-ahead from @baileyckelly as we need to merge some wording into the Terms and Conditions before deploying Hotjar on the portal 
🚧🚧🚧🚧🚧🚧🚧🚧🚧🚧

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -45,8 +45,9 @@
   <script>
     /* track when users exit the site after starting a query */
     window.onbeforeunload = function () {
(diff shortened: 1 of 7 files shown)


baileyckelly:
I updated the privacy policy yesterday. https://kidsfirstdrc.org/policies/ You should be good to go from that perspective.

PR #898, opened 2018-10-22
🚧  💩 demo of fuzzy search

hlminh2000:
requires https://github.com/overture-stack/arranger/pull/460/files


jberube:
@hlminh2000 Can we delete that? Or is it something that could be of use for a future feature?

PR #909, opened 2018-10-30
:ambulance: moves usernsap widget to bottom right

bdolly:


ciaranschutte:
@bdolly does this overlap with social icons on file repository?

rosibaj:
@bdolly  im happy to just move this out once @hlminh2000 reviews!

ciaranschutte:
social-icon-overlap

bdolly:
@ciaranschutte the only other options are in the middle on either side of the screen, or to conditionally set the position per view (which I don't recommend)

allisonheath:
How does it look to the middle on the right?

Footer got moved to increase the data content without scrolling - already finding that helpful.

PR #950, opened 2018-11-20
Donut font and sizing fixes

joneubank:


src/chartkit/components/Donut.js
```diff
@@ -12,7 +12,10 @@ const DonutWrapper = styled('div')`
 `;
 
 const DonutTooltip = ({ id, value }) => (
-  <div style={{ fill: '#404c9a' }} keys={id}>{`${value} Members`}</div>
+  <div style={{ fill: '#404c9a' }} keys={id}>
```
hlminh2000: I know this is here from before but I think these colors are available from `src/chartkit/themes.js`

hlminh2000: or something like that where the colors can be pulled from :P

PR #976, opened 2018-11-23
Feat/gen3 empty studies

hlminh2000:


src/components/UserDashboard/AuthorizedStudies/Gen3Connected.js
```diff
@@ -98,23 +113,29 @@ const Gen3Connected = ({
...
+                const { authorizedFiles, unAuthorizedFiles, consentCodes } = combinedStudyData[
+                  studyId
+                ];
+                return (
+                  <Study
+                    key={studyId}
+                    studyId={studyId}
+                    name={studyShortName}
+                    consentCodes={consentCodes}
+                    authorized={authorizedFiles.length}
```
joneubank: Possible `authorizedFiles` is undefined, would cause error.

src/components/UserDashboard/AuthorizedStudies/Gen3Connected.js
```diff
@@ -98,23 +113,29 @@ const Gen3Connected = ({
...
-                />
-              );
-            })
+            !authorizedStudies.length ? (
+              <Box mt={20}>
+                <NoAuthorizedStudiesMessage />
+              </Box>
+            ) : (
+              authorizedStudies.map(({ studyShortName, id: studyId }) => {
+                const { authorizedFiles, unAuthorizedFiles, consentCodes } = combinedStudyData[
```
joneubank: if no `combinedStudyData[studyId]` then all these vars will be undefined, can't call `.length` on undefined, need check?

joneubank: Variable is built in this file so shouldn't be a problem.

Issue #1009, opened 2018-11-30
FIx casing on Cavatica in card title


rosibaj:
On the Create project card status, this title should be cased `Cavatica` not all caps to be consistent with the rest of the portal 

![Screen Shot 2018-11-30 at 11.36.00 AM.png](https://images.zenhubusercontent.com/5a06111e8a75884b9086f5d8/eb2f9376-718a-493d-817c-ed2cd4d93add)

nyanofthemoon:
@rosibaj Found other instances for this casing issue. Let me know if you would like to address these in different tickets. :)

nyanofthemoon:
@rosibaj Perfect- fixing it everywhere then re-sending!

PR #1015, opened 2018-12-04
WIP: #1009 FIx casing on Cavatica in card title

nyanofthemoon:


nyanofthemoon:
@hlminh2000 @ciaranschutte It's still "Work In Progress"- I believe this is the correct fix, but cannot test locally yet because I need to figure out how to create a Cavatica account to be able to access this part of the UI.

PR #1017, opened 2018-12-05
Feat/slide card react slick

ciaranschutte:


src/components/UserDashboard/CavaticaProjects/index.js
```diff
@@ -50,51 +44,56 @@ const CavaticaProjects = compose(injectState)(({ state: { integrationTokens } })
...
-                  />
-                ),
-              },
-            ]
-          : [{ title: 'CAVATICA Projects', component: cardState => <NotConnected /> }]
-      }
-    />
+    <CavaticaProvider>
+      {({ projects, loading }) => {
+        setUserDimension('dimension6', JSON.stringify(projects));
```
ciaranschutte: @bdolly moved `setUserDimension` to here

src/uikit/Multicard/index.js
```diff
@@ -90,31 +84,38 @@ class Multicard extends Component {
...
+              <Row>
+                {!inactive &&
+                  tabs.map((tab, i) => (
+                    <TabMenu
+                      key={i}
+                      active={i === currentTabIndex}
+                      onClick={() => {
+                        this.setIndex(i);
+                        trackUserInteraction({
+                          category: TRACKING_EVENTS.categories.user.dashboard.widgets._multiCard,
```
ciaranschutte: @bdolly moved `tab` tracking code to here because the headers are now passed in as components

src/components/UserDashboard/CavaticaProjects/index.js
```diff
@@ -50,51 +44,53 @@ const CavaticaProjects = compose(injectState)(({ state: { integrationTokens } })
...
-                    onProjectCreated={onProjectCreationComplete(cardState)}
-                    onProjectCreationCancelled={onProjectCreationCanceled(cardState)}
-                  />
-                ),
-              },
-            ]
-          : [{ title: 'CAVATICA Projects', component: cardState => <NotConnected /> }]
-      }
-    />
+    <CavaticaProvider>
```
ciaranschutte: @hlminh2000 Issue here. All tab content is loaded first time with the slider. When a user creates a new cavatica project we want the `CavaticaProvider` to refresh. Any thoughts on best approach?

Maybe have a `refresh` state at the `CavaticaProjects` level and pass it to `CavaticaProvider`

= `Create` Card =
`onProjectCreationComplete` function can toggle it from the `Create` tab content

= `CavaticaProvider` =
`componentShouldUpdate` if `refresh === true` get data  again

^ Seems a little convoluted

hlminh2000: I'd suggest a solution like the one in gen3's `Gen3UserProvider`. Basically the provider injects a `refresh` function which the consumer can call after its done with side effects :)

PR #1018, opened 2018-12-05
kf-portal-ui#821 Remove 2nd tooltip (icon-controlled-access) from pop…

nyanofthemoon:
…ping up after extended hover


src/components/FileRepo/ui.js
```diff
@@ -114,7 +116,36 @@ export const QuerySharingContainer = styled(Row)`
...
 `;
 
-export const ControlledIcon = props => <ControlledAccessIcon width={12} height={12} {...props} />;
+
+export const ControlledIcon = withTheme((props) => {
+  const initialState = { mouseIsOver: false };
+  const iconContainerRef = React.createRef();
+
+  const didMount = ({ setState }) => {
+    iconContainerRef.current.addEventListener('mouseenter', e => {
```
ciaranschutte: Would it be better to add `onMouseLeave` and `onMouseEnter` as props to the svg icon rather than adding a `ref` and manually doing event listeners seen as React already has those? @hlminh2000 @nyanofthemoon

hlminh2000: @ciaranschutte good point actually, I just never used the `onMouseLeave` and `onMouseEnter` callbacks on react components so didn't think of that, but that is definitely preferable :)

also @nyanofthemoon I think I know why the event listeners attached on svg wasn't working for you :)

src/icons/ControlledAccessIcon.js
```diff
@@ -10,7 +10,7 @@ export default ({ width = 15, height = 15, fill = '#a9adc0', className = '', ...
     height={height}
   >
     <defs />
-    <title>icon-controlled-access</title>
+    { props.hideTitle ? null : <title>icon-controlled-access</title> }
     <path
       fill={fill}
       class="cls-1"
```
ciaranschutte: `className` instead of `class` React will complain :)

nyanofthemoon: Also fixing other files with this problem.

nyanofthemoon:
@ciaranschutte @hlminh2000 Cool! Who merges / how can I merge?

PR #1019, opened 2018-12-05
Entity Pages - Structure and Data only, No Content

joneubank:


joneubank:
In Progress - Need to add API integration.

src/components/EntityPage/File/index.js
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+import { compose } from 'recompose';
+import { injectState } from 'freactal';
+import { withTheme } from 'emotion-theming';
+
+import styled from 'react-emotion';
+import Row from '../../../uikit/Row';
+import Column from '../../../uikit/Column';
```
hlminh2000: I think absolute path reference would be better for uikit :P `uikit/Row` should work

src/components/EntityPage/File/index.js
```diff
@@ -0,0 +1,64 @@
...
+
+const Container = styled(Column)`
+  flex-direction: column;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+`;
+
+const FileEntity = ({ state, effects, theme, fileId, ...props }) => {
+  return (
```
hlminh2000: should we use the react-grid-system here? I feel like here's a page we can benefit from some responsive grid :P

joneubank: Maybe within each section? The entity pages are split into sections with headers then dividing lines. Would be good to put content within them into grid system. But the page itself is just a column of sections

src/components/EntityPage/File/index.js
```diff
@@ -0,0 +1,64 @@
...
+    <Container>
+      <EntityTitleBar>
+        <EntityTitle icon="file" title={fileId} tags={['WGS', 'Aligned Reads']} />
+      </EntityTitleBar>
+      <EntityActionBar>Share Button</EntityActionBar>
+      <EntityContent>
+        <EntityContentSection title="File Properties">
+          <Row style={{ width: '100%' }}>
```
joneubank: The content here is a good candidate for a grid system instead of forced 1 row, 2 column as I wrote it.

Would leave this for whoever is implementing the content I think...

src/components/EntityPage/File/index.js
hlminh2000: 👍

joneubank:
Comments addressed. Arranger data connection added to each page with ugly massive query to get all fields.

Ready to merge, pending final review.

PR #1021, opened 2018-12-06
added billing; still need to fix the css

nyanofthemoon:
Still need to fix the CSS.


src/components/cavatica/CavaticaAddProject.js
```diff
@@ -50,11 +50,26 @@ const AddIcon = styled(PlusIcon)`
   fill: ${({ theme }) => theme.tertiary};
 `;
 
+//           # THIS ONE WE FIXIN
+
+const BillingGroupSelect = styled('select')`
+  ${({ theme }) => theme.select};
+  ${({ theme }) => theme.input};
+`;
```
hlminh2000: This looks like a generic reusable input eh, should we have this exported from uikit? A little bit of continuous refactoring is healthy :) we can of course extend it here to address whatever problem it may have in this context.
Any thoughts on this @ciaranschutte ? :)

nyanofthemoon: @hlminh2000 @ciaranschutte Definitely looks re-usable. Just point me to where and I can update it?

nyanofthemoon: @ciaranschutte Added generic select uikit component

PR #1028, opened 2018-12-10
add explicit dimension attributes to fix rendering in FF

ciaranschutte:


ciaranschutte:
FF needs explicit height and width as attribute not just css styles

PR #1039, opened 2018-12-18
Entity Page -  Info Box

ciaranschutte:


src/uikit/InfoBox/InfoBox/index.js
```diff
@@ -0,0 +1,43 @@
...
+
+const Value = styled('div')`
+  font-weight: 600;
+  font-size: 18px;
+  color: ${({ theme }) => theme.greyScale1};
+  margin-bottom: 2px;
+`;
+
+const Description = styled('div')`
+  font-size: 13px;
```
nyanofthemoon: @ciaranschutte Could we have the font sizes in the theme?

src/uikit/InfoBox/InfoBox/index.js
```diff
@@ -0,0 +1,43 @@
...
+  padding: 20px;
+  flex-basis: 0;
+  flex: 1;
+
+  ${({ styles }) => (styles ? styles : null)};
+`;
+
+const Value = styled('div')`
+  font-weight: 600;
+  font-size: 18px;
```
nyanofthemoon: @ciaranschutte Could we have all the typography (font sizes, weights, etc) in the theme?

ciaranschutte: @nyanofthemoon I think we're moving towards keeping a lot of styling in local styled components; I'm going to leave this here for now

Issue #814, opened 2018-09-28
Make the facets panel width user-resizable


Fgerthoffert:
Add the ability to change the size, or collapse the left facet panel. 

As examples:
- HCMI has the abiltiy to expand (https://hcmi.netlify.com/) 
- GDC has the ability to collapse (https://portal.gdc.cancer.gov/exploration) 

Mockup:
Zeplin: https://zpl.io/bev8YdN
Similar to HCMI, the crosshair icon will show up when the user hovers over the edges of the panels (both facet and action panels).
The title area of the facet panel will change slightly to match the collapsible style of the action panel.

![screen shot 2018-12-04 at 11 14 54 am](https://user-images.githubusercontent.com/18424138/49455854-e18c2700-f7b5-11e8-9dea-b630a60c2c38.png)

rosibaj:
QA - facet can collapse, but still not able to adjust the size of the panel (larger or smaller)

rosibaj:
@nyanofthemoon Vincent wants to go ahead and deploy with JUST the collapse - this looks fine in preview - can you merge to deploy?

nyanofthemoon:
@rosi I merged this into the `vr` branch (preview). Do I need to merge this somewhere else as well? I'll ask in standup this morning.

nyanofthemoon:
@rosibaj @hlminh2000 PR to merge from `vr` to `next` https://github.com/kids-first/kf-portal-ui/pull/1063

rosibaj:
tested on prod - working as expected.  Nice work @nyanofthemoon !

PR #1064, opened 2019-01-10
#814 Make the facets panel width user-resizable

hlminh2000:


src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
```diff
@@ -35,21 +36,41 @@ const AggregationWrapper = styled(Column)`
...
+  transition: all 0.25s;
+  min-width: 37px;
+  flex: inherit;
+  width: ${({ expanded }) => (expanded ? `100%` : `0%`)} !important;
+  overflow: hidden;
+`;
+
+const CollapsibleCustomAggregationsPanel = styled(CustomAggregationsPanel)`
+  display: ${({ expanded }) => (expanded ? `block` : `none`)};
+`;
```
hlminh2000: `CustomAggregationsPanel` will not work with `styled` because it doesn't have a `className` prop that's forwarded to a DOM element, I think what you are looking to do is a `styled('div')` and using it to wrap `CustomAggregationsPanel` in jsx. Might need a `height:100%` as well as it's a new dom element

src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
```diff
@@ -107,35 +129,50 @@ const AggregationSidebar = compose(injectState, withTheme, withApi)(
...
+              />
+            ) : (
+              <CloseActionChevron
+                onClick={() => setExpanded(!expanded)}
+                width={14}
+                fill={theme.secondary}
+              />
+            )}
+          </AggregationHeader>
+          {expanded ? (
```
hlminh2000: so instead of the conditional rendering, we can do:
```
<CollapsibleDiv expanded={expanded}/>
   <CustomAggregationsPanel {...stuff}/>
</CollapsibleDiv>
```
where:
```
const CollapsibleDiv = styled('div')`
   display: ${({ expanded }) => (expanded ? `block` : `none`)};
   height: 100%;
`
```
The div immediately after the Tabs probably needs a `height:100%` too, I think

The cleaner way of going through all this is probably feeding the `expanded` state into `CustomAggregationsPanel` as a `hidden` prop and have it hide itself internally

PR #1066, opened 2019-01-11
🐛 prevents refetch on aggs panel expansion

hlminh2000:
Eventually we should look into introducing client-side caching at the data-fetching layer, but that's a bigger task


src/components/FileRepo/AggregationSidebar/AggregationSidebar.js
```diff
@@ -146,32 +142,31 @@ const AggregationSidebar = compose(
...
-                    InputComponent: FilterInput,
-                  }),
-                },
-              }}
-            />
-          ) : null}
+          <CustomAggregationsPanel
+            {...{
+              ...props,
+              hidden: !expanded,
```
hlminh2000: adding this prop

PR #1087, opened 2019-01-16
:chart_with_upwards_trend: File Entity Page tracking

bdolly:
Hooked up GA events to the page. Swapped the `<ExternalLink>` components for react-router `<Link>` that have hardcoded file repo urls. These are just temp so that we can launch and track engagement while the other entity pages are being built


ciaranschutte:
Note: leaving links back to file repo out of this for now

PR #1100, opened 2019-01-21
Feature/#1056 file entity page responsiveness

ciaranschutte:


src/components/EntityPage/File/index.js
```diff
@@ -330,12 +327,7 @@ const FileEntity = ({ api, fileId }) => {
               <EntityContent>
                 <EntityContentSection title="File Properties">
                   <Row style={{ width: '100%' }}>
```
hlminh2000: completely just a nit pick :P can we do `className={css'...'}` to stick with emotion? :)

src/uikit/InfoBox/InfoBox/index.js
```diff
@@ -1,19 +1,18 @@
 import React from 'react';
-import styled from 'react-emotion';
+import styled, { css } from 'react-emotion';
```
hlminh2000: unused `css`

src/uikit/SummaryTable/index.js
```diff
@@ -1,52 +1,101 @@
...
-      {rows.map((row, i) => (
-        <TableRow index={i} key={i}>
-          <TableTitle>{row.title}</TableTitle>
-          <TableCell>{row.summary}</TableCell>
-        </TableRow>
-      ))}
-    </tbody>
-  </Table>
-);
+class SummaryTable extends React.Component {
```
hlminh2000: I think this component is easier to understand as a function, so the `rows` is just fed directly down :P should be a quick switch, any harm? :)

PR #1130, opened 2019-01-22
Storybook cleanup

hlminh2000:


.storybook/webpack.config.js
```diff
@@ -0,0 +1,9 @@
+const path = require('path');
+
+module.exports = (config, env) => {
+  config.module.rules = [
+    ...config.module.rules,
+    { test: /\.(gif|svg|jpg|png)$/, loader: 'file-loader' },
+  ];
+  return config;
+};
```
hlminh2000: this webpack config is only used by storybook. had to include `file-loader` manually and update its version so images can be loaded properly.

README.md
```diff
@@ -36,18 +35,30 @@ Copy `.env.schema` to `.env.local` and configure it with appropriate endpoints.
...
+  - feat (feature)
+  - bug (bug fix)
+  - docs (documentation)
+  - refactor
+  - test (when adding missing tests)
+  - chore (maintain)
 
   Once development is complete for the scope defined by the supporting branch, a pull request can be made for the `next` branch for code review.
 
+- ### Component Development
```
joneubank: This is very helpful. Do we also want to put in a note about where storybook code lives and/or an example for starting a new component? - Probably optional for us but helpful?

src/common/injectGlobals.js
```diff
@@ -1,7 +1,9 @@
-import 'index.css';
```
hlminh2000: having this here seems weird, moved to `src/index.js`

src/common/injectGlobals.js
```diff
@@ -1,7 +1,9 @@
...
 import queryString from 'querystring';
 
-localStorage.setItem('debug', process.env.REACT_APP_DEBUG || ''); // manually set because CRA doesn't allow arbitrary env variable names.
+const getApplicationEnvVar = envVarName =>
```
joneubank: This will choose the REACT_APP_* variable if available instead of the STORYBOOK_* variable. Is this intended? Do we want a command line arg in the run storybook script that could be used to select storybook instead?

hlminh2000: That was my intention. I didn't think of extending the script, no objection on such feature other than it's extra logic which I personally don't think we need. If you think that's a better approach, LMK ;)

README.md
hlminh2000: yes, can certainly add! thanks! :)

src/common/injectGlobals.js
```diff
@@ -13,70 +15,73 @@ if (egoApiOverride) {
...
+export const hotjarId = getApplicationEnvVar('HOTJAR_ID');
 
-export const gaTrackingID: string = process.env.REACT_APP_GA_TRACKING_ID;
+export const gaTrackingID: string = getApplicationEnvVar('GA_TRACKING_ID');
 
-export const defaultRedirectUri = process.env.REACT_APP_DEFAULT_REDIRECT_URI || '';
+export const defaultRedirectUri = getApplicationEnvVar('DEFAULT_REDIRECT_URI') || '';
 
-export const allRedirectUris = (process.env.REACT_APP_LOGIN_REDIRECT_NOT_REQUIRED || '')
+export const allRedirectUris = getApplicationEnvVar('LOGIN_REDIRECT_NOT_REQUIRED' || '')
```
hlminh2000: oops, bug here 📦

hlminh2000: fixed!

src/common/injectGlobals.js
joneubank: Can merge this without concern since the normal params take precedence... if you want to run your storybook with different values you could end up with conflict which would be frustrating. If that happens we can deal with it then.

PR #1135, opened 2019-01-24
Cohort Builder - Page Structure

ciaranschutte:
The page is still a work in progress but this is the barest bones skeleton.
Will hopefully remove blockers :)

### DIFF
--- a/src/components/Header/index.js
+++ b/src/components/Header/index.js
@@ -63,11 +63,6 @@ const Header = ({
                     </NavLink>
                   </li>
                   <li>
-                    <NavLink currentPathName={currentPathName} to={`/cohort-builder`}>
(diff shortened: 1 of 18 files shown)


src/components/Header/index.js
```diff
@@ -142,4 +141,9 @@ const Header = ({
   );
 };
 
-export default compose(injectState, withTheme, withRouter, withApi)(Header);
+export default compose(
+  injectState,
+  withTheme,
+  withRouter,
+  withApi,
+)(Header);
```
hlminh2000: I'm guessing everything here is just prettier... can you confirm @ciaranschutte? :)

ciaranschutte: Confirmed :)

PR #1139, opened 2019-01-28
Feature - Cohort Builder Styling

ciaranschutte:


src/icons/AllAppsMenuIcon.js
```diff
@@ -1,57 +1,64 @@
...
-    width={width}
-    height={height}
-    css={`
-      margin-right: 10px;
-    `}
-    {...props}
   >
-    <defs />
+    <defs >
+      <style>.cls-1{fill: ${fill}</style>      </defs>
```
hlminh2000: I'm a bit paranoid of using class name here, simply because of potential leaks. I think all the svg we get come with that `cls-1` class so if we allow that classname, they can mess with each other. Please correct me if I'm plain wrong, but a solution could be to put the fill as props for each of the `path`? :)

src/icons/DemographicIcon.js
```diff
@@ -0,0 +1,10 @@
+import { BaseSvg } from 'icons';
+
+export default ({ fill = '#a9adc0', width = 14, height = 17, marginRight = 4, ...props }) =>
+  BaseSvg({
+    svg: `<svg id="Isolation_Mode" data-name="Isolation Mode" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 37 45.49"><defs><style>.cls-1{fill:${fill};fill-rule:evenodd;}</style></defs><title>icon-demographic</title><path class="cls-1" d="M11.24,31.52c0,4.77-2.78,11.06.23,13.57,2.32,1.93,4.63-3.48,6.94-7.45,2.64-4.53,4-3.57,9.5-4.07,12.8-1.17,10.17-4.61,3.77-5.68-2.38-.4-5.87-.92-7.48-4.43-1.53-3.32.1-6.47,2.65-10.09,3.46-4.9,4.45-7.72-3.12-3.42-7.8,4.44-8.22,5.58-16.41,1.29-11-5.76-7.65.1-2.63,5A21.67,21.67,0,0,1,11.24,31.52Z"/><path class="cls-1" d="M11.53,10.42c-1.91-2.1-1.57-6.64.56-8.58,2.72-2.47,5.12-2.19,6.61-.7C23.88,6.34,15.7,15,11.53,10.42Z"/></svg>`,
```
hlminh2000: `class="cls-1"`

src/icons/FilesIcon.js
```diff
@@ -0,0 +1,10 @@
+import { BaseSvg } from 'icons';
+
+export default ({ fill = '#a9adc0', width = 9, height = 12, marginRight = 6, ...props }) =>
+  BaseSvg({
+    svg: `<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 46 56"><defs><style>.cls-1{fill:${fill};}</style></defs><title>icon-files</title><path class="cls-1" d="M44.8,12H34V1.2ZM32,16a2,2,0,0,1-2-2V0H2A2,2,0,0,0,0,2V54a2,2,0,0,0,2,2H44a2,2,0,0,0,2-2V16Z"/></svg>`,
```
hlminh2000: `class="cls-1"`

src/icons/TableViewIcon.js
```diff
@@ -0,0 +1,10 @@
+import { BaseSvg } from 'icons';
+
+export default ({ fill = '#a6278f', width = 12, height = 12, marginRight, ...props }) =>
+  BaseSvg({
+    svg: `<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 84.99 82"><defs><style>.cls-1{fill:${fill}</style></defs><title>icon-table-view</title><path class="cls-1" d="M0,52H85V30H0ZM0,82H85V60H0ZM0,0V22H85V0Z"/></svg>`,
```
hlminh2000: `class="cls-1"`

src/uikit/ViewLink.js
```diff
@@ -1,21 +1,22 @@
...
-const Link = styled('div')`
-  ${({ theme }) => console.log('theme', theme)}
+const Link = styled(Row)`
+  padding-bottom: 5px;
   font-family: Montserrat;
   font-size: 14px;
   font-weight: 500;
   color: #cc3399;
-  border-bottom: ${({ active }) => (active ? '1px solid #cc3399' : null)};
+  border-bottom: ${({ active }) => (active ? '2px solid #cc3399' : null)};
```
hlminh2000: magic color :P

src/icons/AllAppsMenuIcon.js
ciaranschutte: For reference :)
 
Once we keep wrapping our SVGs in our BaseSVG class we are good because it turns it into an image with a data URL source

ciaranschutte:
@hlminh2000 Added theming :)

PR #1145, opened 2019-02-01
requests by kim and prettier fixes

nyanofthemoon:


src/chartkit/components/Pie.js
```diff
@@ -41,10 +64,17 @@ class Pie extends Component {
     }));
 
     return (
-     <PieWrapper style={this.props.style}>
-       {this.props.title ? <PieTitle>{this.props.title}</PieTitle> : null }
-       <ResponsivePie {...this.props} data={colorData} colorBy={data => data.color} />
-     </PieWrapper>
+      <PieWrapper style={this.props.style}>
+        {this.props.title ? <PieTitle>{this.props.title}</PieTitle> : null}
```
ciaranschutte: Thoughts on pulling `title` and `style` out above it it with `const {style, tittle} = this.props` to clean it up a bit?

ciaranschutte: @nyanofthemoon

src/chartkit/components/Pie.js
```diff
@@ -19,17 +19,40 @@ const PieTitle = styled('div')`
   margin-bottom: 5px;
 `;
 
+const StyleTooltip = styled('Tooltip')`
```
hlminh2000: I don't think the style is applying, we'll need `className` passed to the DOM element in the `Tooltip` component for it to apply. Also I think it should be `styled(Tooltip)` (`Tooltip` as component reference, not string)

src/chartkit/components/Pie.js
hlminh2000: I'm not opposed to neither for class components. Though I personally think functional components remove these kind of noise, but it's mainly a personal preference :P

hlminh2000: I'm not opposed to neither for class components. Though I personally think functional components remove these kind of noise, but it's mainly a personal preference :P

nyanofthemoon: Not opposed, just wanted to do the cleanup after plugging in the data provider; there might be additional things to cleanup.

src/chartkit/components/Pie.js
nyanofthemoon: @hlminh2000 Works for me using the component name as a string.

nyanofthemoon: screen shot 2019-02-04 at 11 53 44 am

hlminh2000:
oh the conflict thought... :P

PR #1146, opened 2019-02-01
chart + prettier

nyanofthemoon:


src/components/CohortBuilder/Summary/index.js
```diff
@@ -2,20 +2,40 @@ import React from 'react';
...
+      <div>{`${participants.toLocaleString()} Participant${participants > 1 ? 's' : ''}`}</div>
+    </div>
+  );
+};
+
+const sortDescParticipant = (a, b) => {
+  const aTotal = a.probands + a.familyMembers;
+  const bTotal = b.probands + b.familyMembers;
+  return aTotal <= bTotal ? -1 : 1;
+};
```
hlminh2000: @ciaranschutte I missed this in review way back, but just noticed that we're using `Array.sort` in the `HorizontalBar`. `Array.sort` actually does an in-place sort so, we prefer to use `lodash`'s version :P It takes a different callback signature though so will take some modifications to stuff currently using it. No need to worry about this now, just mentioning so we're aware ;)

src/components/CohortBuilder/Categories/Category.js
```diff
@@ -14,7 +14,36 @@ const Container = styled(Column)`
...
+  }
+`;
+
+const ItemWrapper = styled('div')`
+  display: flex;
+  padding: 17px 10px 17px 23px;
+  font-size: 12px;
+  color: #343434;
+  font-weight: 500;
+>>>>>>> 53be2c1397f94f51878971172c673ed4f75cb4d5
```
ciaranschutte: Git hash in code

PR #1155, opened 2019-02-04
fixed #1132 - Experimental Strategy summary being empty

### LINKED ISSUE #1132: Update Experimental Strategy field on File Entity PAge

jberube:


ciaranschutte:
@jberube @jberube I don't we need to worry about `data.experiment_strategies` seen as it's `data.sequencing_experiments` we're targeting.

Also did you happen to try something like: 
```
pickData(data, 'sequencing_experiments.hits.edges', experiments => _.uniq(experiments.filter(exp => exp).map(exp => exp.node.experiment_strategy).join(', '))

```

jberube:
Using `pickData` can cause duplicates as `_.uniq` would be run independently for each `sequencing_experiments.hits.edges` instance, causing duplicates.

src/components/EntityPage/File/fileProperties.js
```diff
@@ -10,6 +11,18 @@ export const toFilePropertiesSummary = data => {
   const study = participants.study;
   const biospecimens = participants.biospecimens.hits.edges[0].node;
 
+  // it could have been easier (see below), but `data.experiment_strategies` is sometimes empty,
+  //  even if data.sequencing_experiments.hits.edges is ont empty for the given file 😕
+  // const experimentalStrategies =
+  //   _.uniq(_.get(data, 'experiment_strategies').filter(datum => !!datum)).join(', ') || '--';
```
hlminh2000: I think we can lose this comment, the `data.experiment_strategies` field is to be removed anyway :)

PR #1170, opened 2019-02-05
added chart

nyanofthemoon:


ciaranschutte:
@nyanofthemoon probably worth putting `TextBugWrapper` into it's own style file to import now seen as it's gonna be used by all chart text ever. Apart from that looks good

nyanofthemoon:
@ciaranschutte Done.

PR #1172, opened 2019-02-06
fixes 1094 - sets default value of 'Proband' correctly.

jberube:
…Also, better check for  null value.

`pickData` does not invoke the transform when the field is absent or null (`_.get` returns null).

```
export const pickData = (data, valuePath, transform = x => x) => {
  const selectedData = _.get(data, valuePath, null);
  return selectedData === null ? '--' : transform(selectedData);
};
```

Would you prefer to invoke the transforms even with null values?

```
export const pickData = (data, valuePath, transform = x => x) => {
  const selectedData = _.get(data, valuePath, null);
  return transform(selectedData);
};
```

Pro: each field can transform a null value.
Con: more risk of inserting a regression, maybe some minor refactor on some usage of `pickData`.

Another option is to return `--` by default if no `transform` is provided:

```
export const pickData = (data, valuePath, transform) => {
  const selectedData = _.get(data, valuePath, null);
  if (transform) {
    return transform(selectedData);
  }
  return selectedData === null ? '--' : selectedData;
};
```

### DIFF
--- a/src/components/EntityPage/File/index.js
+++ b/src/components/EntityPage/File/index.js
@@ -38,7 +38,7 @@ import {
 } from './experimentalStrategies';
 
 import { toFilePropertiesSummary } from './fileProperties';
-import { toSequencingReadProperties } from './sequencingProperties';
+import { hasSequencingReadProperties, toSequencingReadProperties } from './sequencingProperties';
 
 import CavaticaAnalyse from './CavaticaAnalyse';
 import Download from './Download';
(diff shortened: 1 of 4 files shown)


ciaranschutte:
@jberube transforming even with null values means all transforms will have to check for nulls, which I was trying avoid by having a pickData function. So maybe 2nd option is good. Whatever you consider best.
@hlminh2000 thoughts?

jberube:
I'm done with that one.

PR #1177, opened 2019-02-06
styling

nyanofthemoon:


src/chartkit/components/HorizontalBar.js
```diff
@@ -52,7 +55,8 @@ class HorizontalBar extends Component {
   }
 
   onMouseEnter(data, e) {
-    e.target.style.cursor = 'pointer';
+    //@NOTE Re-enable once graph is clickable
```
ciaranschutte: V important. User Dashboard has charts with pointer for clickable currently. This breaks it.

nyanofthemoon: @ciaranschutte Was requested to remove this for now by @rosibaj

nyanofthemoon: @ciaranschutte No "easy" way to remove temporarily, so putting back.

rosibaj: @ciaranschutte @nyanofthemoon Not sure on code here, but its _very_ important to have the pointer where needed, but not where not needed.  Can we work out a way for that to happen?

PR #1178, opened 2019-02-07
Feature - Demographic Chat Data Provider

ciaranschutte:
One thing to note.

Looks like it makes more sense to pass SQON to QueriesResolver instead of to the transform functions.
I will make this change in https://app.zenhub.com/workspaces/oicr-sprints-59a846e9d6fb732a00631122/issues/kids-first/kf-portal-ui/1157 work

### DIFF
--- a/src/components/CohortBuilder/QueriesResolver.js
+++ b/src/components/CohortBuilder/QueriesResolver.js
@@ -8,7 +8,7 @@ const QueriesResolver = ({ children, queries, api }) => (
     initialState={{ data: null, isLoading: true, error: null }}
     didMount={({ setState }) => {
       const { query, variables, transform = x => x } = queries[0];
-      console.log('query', query, 'variables', variables);
+
       graphql(api)({
         query,
         variables,
--- a/src/components/CohortBuilder/QueriesResolver.js
+++ b/src/components/CohortBuilder/QueriesResolver.js
@@ -13,16 +13,19 @@ const QueriesResolver = ({ children, queries, api }) => (
         query,
         variables,
       })
-        .then(data => {
-          console.log('data', data);
-          return data;
-        })
         .then(data => transform(data))
+        .then(d => {
+          console.log('transform', d);
+          return d;
+        })
         .then(data => setState({ data: data, isLoading: false }))
         .catch(err => setState({ isLoading: false, error: err }));
     }}
   >
-    {({ state }) => children(state)}
+    {({ state }) => {
+      console.log('state', state);
+      return children(state);
+    }}
   </Component>
 );
 
@@ -33,7 +36,7 @@ QueriesResolver.propTypes = {
   queries: PropTypes.arrayOf(
     PropTypes.shape({
       query: PropTypes.string.isRequired,
-      variables: PropTypes.string.isRequired,
+      variables: PropTypes.isRequired,
       transform: PropTypes.func,
     }),
   ).isRequired,
--- a/src/components/CohortBuilder/QueriesResolver.js
+++ b/src/components/CohortBuilder/QueriesResolver.js
@@ -14,18 +14,11 @@ const QueriesResolver = ({ children, queries, api }) => (
         variables,
       })
         .then(data => transform(data))
-        .then(d => {
-          console.log('transform', d);
-          return d;
-        })
         .then(data => setState({ data: data, isLoading: false }))
         .catch(err => setState({ isLoading: false, error: err }));
     }}
   >
-    {({ state }) => {
-      console.log('state', state);
-      return children(state);
-    }}
+    {({ state }) => children(state)}
   </Component>
 );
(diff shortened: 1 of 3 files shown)


src/components/CohortBuilder/QueriesResolver.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Component from 'react-component-component';
+import graphql from 'services/arranger';
+
+const QueriesResolver = ({ children, queries, api }) => (
+  <Component
+    initialState={{ data: null, isLoading: true, error: null }}
+    didMount={({ setState }) => {
+      const { query, variables, transform = x => x } = queries[0];
```
nyanofthemoon: @ciaranschutte Why queries[0]?

nyanofthemoon: Oic, you will pass on all of the pie queries as an array,

src/components/CohortBuilder/Summary/index.js
```diff
@@ -39,118 +39,95 @@ const CardSlot = styled(Card)`
...
-      </Row>
-    </Col>
-    <Col sm={12} md={3} lg={3}>
-      <LongCard title="Phenotypes">
-        <pre>{JSON.stringify(sqon, null, 2)}</pre>
-      </LongCard>
-    </Col>
-  </Row>
+const Summary = ({ theme, sqon, api }) => (
+  <QueriesResolver api={api} sqon={sqon} queries={[demographicQuery({ ...{ sqon } })]}>
```
nyanofthemoon: @ciaranschutte I don't think this needs to be another HOC, but that is just my preference.

src/components/CohortBuilder/Summary/index.js
```diff
@@ -39,118 +39,95 @@ const CardSlot = styled(Card)`
...
+    {({ loading, data }) => {
+      return loading || !data ? (
```
nyanofthemoon: @ciaranschutte Does this mean that the user will have to wait for all queries to be resolved before seeing any results? Are we including error results as part of another ticket?

ciaranschutte: @nyanofthemoon Yes. It means all queries will need to be resolved. I can include include error results in the global sqon ticket: https://app.zenhub.com/workspaces/oicr-sprints-59a846e9d6fb732a00631122/issues/kids-first/kf-portal-ui/1157

src/components/CohortBuilder/Summary/index.js
ciaranschutte: @nyanofthemoon After talking with @hlminh2000 HOC is probably best for now.
So we have a state provider for the SQONs to the page
We also need to provide data to the charts - this is what the `QueriesResolver` is for

src/components/CohortBuilder/QueriesResolver.js
ciaranschutte: @nyanofthemoon Exactly. `queries[0]` to just have this one working.
It will be passed an array something like: 
``` javascript
[
  {
    query: `{
      file {
        aggregations {
          kf_id {
            buckets {
              key
            }
          }
        }
      }
    }`,
    variables: {
      sqon: sqon
    }
  },
  {
    query: `{
      file {
        aggregations {
          kf_id {
            buckets {
              key
            }
          }
        }
      }
    }`,
    variables: {
      sqon: sqon
    }
  },
```

This is part of https://app.zenhub.com/workspaces/oicr-sprints-59a846e9d6fb732a00631122/issues/kids-first/kf-portal-ui/1157

src/components/CohortBuilder/QueriesResolver.js
```diff
@@ -0,0 +1,36 @@
...
+
+      graphql(api)({
```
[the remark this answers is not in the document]
hlminh2000: on another note, this whole logic here needs to be replicated when the sqon changes, so it's one of the rare occasions I would recommend class components (this is actually also a great use-case of the new React Hook's `useEffect`, but for now a class component with a `componentWillReceiveProps` will do ;)

src/components/CohortBuilder/QueriesResolver.js
hlminh2000: yes, that's the idea :) all the queries will be batched into one single ajax request

src/components/CohortBuilder/Summary/index.js
hlminh2000: The state provider only provides the sqon, `QueriesResolver` resolves that sqon into actual data from the API.

hlminh2000: @jberube yup, I mentioned a solution to that problem below :) the implementation of `QueriesResolver` can change but you got the idea ;)

PR #1207, opened 2019-02-15
🐛 converts demographic chart to use aggregations

hlminh2000:
aggregations filter themselves


src/components/CohortBuilder/Summary/DemographicChart.js
```diff
@@ -46,43 +47,50 @@ const DemographicChart = ({ data, theme }) => (
...
-              family_compositions {
-                hits {
-                  edges {
-                    node {
-                      composition
-                    }
-                  }
-                }
-              }
+  query: gql`
```
hlminh2000: I have no idea why we never used this, just one of those things that should have been there since the beginning, but not somehow...

Issue #1196, opened 2019-02-12
Card Styling: Remove Participants from Study Card


rosibaj:
- [x] Remove the `Participants (xxx)` from the title of the Study Breakdown Card
![image.png](https://images.zenhubusercontent.com/5a06111e8a75884b9086f5d8/d956aa98-faf9-40ea-a813-86e183a4ea35)

ciaranschutte:
@rosibaj @hlminh2000 this is complete

PR #1209, opened 2019-02-15
💄Fixes the outstanding cards layout issues

hlminh2000:
- [x] (#1196 ) For the studies chart, we can remove the “# Participants” stat from the title bar as it’s redundant with the count above it.
- [x] Consistent card heights
preview: https://5c66cb9b1094df0008eb3e99--kf-portal-storybook.netlify.com/?selectedKind=Storybook%20ego%20login&selectedStory=login&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel


src/uikit/Card/index.js
```diff
@@ -12,15 +13,27 @@ const Card = ({
...
+  Header: PropTypes.node,
+  Content: PropTypes.func,
+  children: PropTypes.node,
+  className: PropTypes.string,
+  scrollable: PropTypes.bool,
+  title: PropTypes.node,
+  inactive: PropTypes.bool,
+  showHeader: PropTypes.bool,
+};
+
```
hlminh2000: @jberube I hope this gives you a smile :)

jberube: Yes! :smile_cat: 

None of the props are marked as `.isRequired`, are thay all optional?

What happen if you do that?
```
<Card />
```

Marking the the required props as so and providing defaults otherwise, when necessary, would make it a 10 to me.

src/components/CohortBuilder/Summary/DemographicChart.js
```diff
@@ -4,19 +4,21 @@ import { withTheme } from 'emotion-theming';
...
+  position: absolute;
+  left: 0px;
+  right: 0px;
+  top: 0px;
+  bottom: 0px;
 `;
 
 const DemographicChart = ({ data, theme }) => (
-  <CardSlotPies>
+  <Container>
```
jberube: Since you named the container of the bar chart `BarChartContainer`, I suggest that you remain constant by naming this one `<PieChartContainer>`.

src/components/CohortBuilder/Summary/index.js
```diff
@@ -96,51 +101,57 @@ const Summary = ({ theme, sqon, api }) => (
         <Row nogutter> no data</Row>
       ) : (
         <Row nogutter>
-          <Col sm={12} md={12} lg={9}>
+          <Col sm={12} md={12} lg={12} xl={9}>
```
jberube: I think you can omit some of those.

If react-grid-system really works like bootstrap, those parameters are to be interpretet this way:
`xs={12}` means `12 columns when xs _or bigger_, unless overrided by a larger screenClass
`md={6}` means `9 columns when sm _or bigger_, unless overrided by a larger screenClass

So the above example would yoeld:
xs=12, sm=12, md=6, lg=6, xl=6

hlminh2000: I'll try that out, thanks! :)

src/uikit/Card/index.js
hlminh2000:
```
<Card />
```
would work just fine, would just get an empty card :P
screen shot 2019-02-15 at 10 01 19 am

ciaranschutte:
@hlminh2000 @jberube  Can we officially have a "Julien -10" as our gold standard? :D

src/uikit/Card/index.js
jberube: :+1:

Issue #1180, opened 2019-02-07
Table actions: columns dropdown button usability


kcullion:
Clicking on the columns button should open and close the dropdown. Currently it opens it but doesn't close it. 
(see file repository table for correct functionality)
![Screen Shot 2019-02-07 at 10.58.18 AM.png](https://images.zenhubusercontent.com/5a5f5bcd4b5806bc2bc31a86/bab02408-8276-4506-8413-6362fee1bd5e)

jberube:
Fixed as part of #1193

rosibaj:
Tested on VR - this issue is resolved with some other commits.  I can close the columns dropdown by clicking on columns.

PR #1215, opened 2019-02-19
Feature/1193 participant table

jberube:
Also provide a fix to #1180


src/components/FileRepo/arrangerStyle.js
```diff
@@ -198,7 +198,6 @@ export default () => css`
     cursor: pointer;
     display: flex;
     align-items: center;
-    justify-content: flex-end;
```
ciaranschutte: @jberube looking great so far. Is this affecting CohortBuilder styles? Might be better to change a `CohortBuilder` style than to change `FileRepo` style?

src/App.js
```diff
@@ -17,6 +17,7 @@ import Join from 'components/Join';
 import LoginPage from 'components/LoginPage';
 import FileEntity from './components/EntityPage/File';
 import ParticipantEntity from './components/EntityPage/Participant';
+import CohortBuilder from './components/CohortBuilder';
```
hlminh2000: 🔥 :)

Ciaran is looking at hooking this up with lazy loading actually :) just wanna remove potential conflicts

src/components/CohortBuilder/Table/ParticipantsTable.js
```diff
@@ -0,0 +1,83 @@
...
+          <div className={`showMore-wrapper ${collapsed ? 'more' : 'less'}`}>
+            {collapsed ? ' More' : 'Less'}
+          </div>
+        </div>
+      ) : null}
+    </div>
+  );
+});
+
+const StyledCollapsibleMultiLineCell = styled(CollapsibleMultiLineCell)`
```
hlminh2000: style might not be applied, needs to pass `className` prop down to a dom element in `CollapsibleMultiLineCell` :)

src/components/CohortBuilder/Table/TableErrorView.js
```diff
@@ -0,0 +1,39 @@
...
+  <div>
+    {`Try refreshing the page and if the error persists, `}
+    <ExternalLink href={`${kfWebRoot}/contact/`} hasExternalIcon={false}>
+      contact us
+    </ExternalLink>
+  </div>
+);
+
+const TableErrorView = ({ error }) => {
+  const message = devDebug ? (
```
hlminh2000: 👍

src/uikit/DataTable/BaseDataTable.js
```diff
@@ -69,19 +70,28 @@ const BaseDataTable = ({
...
+BaseDataTable.propTypes = {
+  columns: PropTypes.arrayOf(
+    PropTypes.shape({
+      Header: PropTypes.string.isRequired,
+      accessor: PropTypes.string.isRequired,
+      Cell: PropTypes.oneOfType([PropTypes.string, PropTypes.func, PropTypes.element]),
+    }),
+  ).isRequired,
+  data: PropTypes.array.isRequired,
+};
```
hlminh2000: 👍

src/components/FileRepo/arrangerStyle.js
jberube: The same style is inherited by the only component that uses it.
As a rule of thumb, I suggest that everything that concerns the positioning of a component should be kept outside of that component's styles. Not all usages of a component are aligned the same way.

src/uikit/DataTable/ToolbarButtons/ColumnFilter.js
```diff
@@ -7,10 +7,14 @@ class ColumnFilter extends Component {
...
-      isOpen: type === Downshift.stateChangeTypes.mouseUp ? false : true,
-    });
+    const { type, isOpen } = changes;
+    if (typeof isOpen !== 'undefined' && type === Downshift.stateChangeTypes.clickButton) {
+      this.setState({ isOpen });
+    }
+  };
+
+  handleClose = () => {
+    this.setState({ isOpen: false });
```
hlminh2000: lgtm but any implication of this change on existing components that comes to mind @ciaranschutte? It'd be great if you could check this out in the portal :)

src/components/FileRepo/arrangerStyle.js
hlminh2000: totally agree @jberube, is this affecting the new feature though? I am just a bit hesitant of changes to existing working features right now. If this is affecting new features, my bigger concern is the fact that this style here is not scoped to the FileRepo only, but addressing that is probably a bigger task.

src/uikit/DataTable/ToolbarButtons/ColumnFilter.js
ciaranschutte: Offhand I think this actually fixes a bug in the other place it's used on `File Entity`. Having some environment issues right now, but will check asap

jberube: This is a fix to #1180 , done on this branch since it's the only way I was able to reproduce.

I should have mentioned that in the description of this PR. :thinking:

src/components/FileRepo/arrangerStyle.js
jberube: I've checked the FileRepo afterward, and it didn't changed there, since it is also set this way:
```
.aggregation-card .showMore-wrapper {
    display: -ms-flexbox;
    display: flex;
    -ms-flex-pack: end;
    justify-content: flex-end;
}
```

src/components/CohortBuilder/Table/ParticipantsTable.js
jberube: You're right _and_ most of that is dead code, but some of it is relevant.
I'll clean that.

src/uikit/DataTable/ToolbarButtons/ColumnFilter.js
hlminh2000: 👍 thank you! :)

src/components/CohortBuilder/Table/ParticipantsTable.js
hlminh2000: which ever DOM element this is supposed to apply to, I think we'd prefer just use `styled` on that element :P the whole `className` can get complex to trace down. Sometimes it's good for re-usable components to allow overrides though :P

src/components/CohortBuilder/Results.js
```diff
@@ -7,7 +7,7 @@ import Row from 'uikit/Row';
 import ViewLink from 'uikit/ViewLink';
 import styled from 'react-emotion';
 import { H2 } from 'uikit/Headings';
-import Table from './Table';
+import ParticipantsTableView from './Table';
```
hlminh2000: would it be too much if we renamed `/Table` to `/ParticipantsTableView`? :)

jberube: The filename? Np.

PR #1229, opened 2019-02-20
bug/1224 - set the correct icon in the page menu for "Explore Data"

jberube:
Set that icon (selected page) :
image


jberube:
> So because of some browser compatibility bug we try to wrap svgs in a function like in `kf-portal-ui/src/icons/CommunityIcon.js`. Not all icons have been wrapped up, but let's try to get the new ones wrapped up :)

@hlminh2000  The problem in my case for that approach is that the icon switches color when the menu choice is hovered, so I've used the 'fill: currentColor` svg property to accomplish that, just like the two other menu choice do, which is way more convenient than the alternative.

Do you know which browsers in particular causes an issue? Maybe we can test if the way I did it really causes an issue.

ciaranschutte:
@jberube I'm not sure about the issues but what I have been doing is passing a `fill` to the wrapped icon.
Check out the `icons/BiospecimentIcon.js`, it is in use like so ` <BiospecimenIcon fill={theme.biospecimenOrange} />`

PR #1233, opened 2019-02-21
🐛uses family_id instead of family.family_id

hlminh2000:
Some of the indices under `participant_centric` in Elasticsearch do not have the `family.family_id` field in their mappings, resulting in inconsistent GraphQL API generated by arranger. 

While this is a symptom of a deeper back-end issue, this UI fix ensures the table will always load as the `family_id` at the root level of participant is confirmed to be the reliable field.

@rosibaj FYI :)


rosibaj:
thanks for the heads-up @hlminh2000 - will check this out and see why its happening!

PR #1239, opened 2019-02-22
Feature/cb style results header

ciaranschutte:
CSS Fixes.

Moved all the card/padding etc into the singular chart files.
Moved a lot of styling to `ui.js`

Found out you can do this in Emotion, pretty cool: 
``` javascript

const CohortCardHeader = styled(CardHeader)`
  font-size: 16px;

  ${Badge} {
    line-height: 20px;
    height: 20px;
    min-width: 20px;
  }
`;
```


ciaranschutte:
@hlminh2000 Moved layout back to index. all good apart from that? :)

PR #1257, opened 2019-02-27
♻️ removes the spread props in QueriesResolver

hlminh2000:
I don't think this is necessary, but please correct me if I'm wrong :)

Removing it now so we don't accidentally rely on it


ciaranschutte:
@hlminh2000 yeah I think we're good taking this out :)

Issue #1226, opened 2019-02-20
Bug: Footer in cohort builder


kcullion:
The footer should be fixed to the bottom of the screen (like in the dashboard).

<img width="1615" alt="screen shot 2019-02-20 at 12 40 09 pm" src="https://user-images.githubusercontent.com/18424138/53116899-ba710600-3517-11e9-8430-9ee5d0ed9833.png">

<img width="1615" alt="screen shot 2019-02-26 at 2 30 20 pm" src="https://user-images.githubusercontent.com/15824252/53440743-429d5280-39d3-11e9-8f86-8dbcd0d8de36.png">

rosibaj:
@kcullion are you still able to reproduce this on your monitor size?

kcullion:
@rosibaj yes I still see it on my monitor (the space below the footer)

![screen shot 2019-02-25 at 4 39 16 pm](https://user-images.githubusercontent.com/18424138/53370548-fd6a1980-391b-11e9-9298-0d60d4b1d1f6.png)

rosibaj:
@kcullion Can you QA this one on your screen size?

kcullion:
@rosibaj , @jaouad-benassila I still see space under the footer.  It seems ok on the dashboard though (stuck to the bottom of the screen)
![screen shot 2019-03-04 at 9 27 21 am](https://user-images.githubusercontent.com/18424138/53739598-24ff3b80-3e60-11e9-8e62-9211dcfacdb7.png)

jaouad-benassila:
@kcullion , @rosibaj it works for me and I can't reproduce it
![footer](https://user-images.githubusercontent.com/47750690/53745851-df496f80-3e6d-11e9-8b69-a2ef8327fcc1.png)

kcullion:
@jaouad-benassila does anyone there have a large monitor?
@rosibaj should we ask @ciaranschutte to look at this since I sit by him and we can test easily?

kcullion:
@jaouad-benassila @ciaranschutte does it have something to do with the treatment of the div just above the footer? or with positioning? 

They are treated differently when I inspect...

![screen shot 2019-03-04 at 11 15 17 am](https://user-images.githubusercontent.com/18424138/53746355-018fbd00-3e6f-11e9-9c39-bf1bdc100024.png)
[...]

PR #1259, opened 2019-02-27
Bug/footer cohort builder

jaouad-benassila:
Bug: Footer in cohort builder #1226


src/components/Page.js
```diff
@@ -24,21 +24,23 @@ const FloatFooterPageComponentWrapper = styled('div')`
...
+  </Container >
 );
 
 export const FixedFooterPage = ({ Head = Header, Foot = Footer, Component, ...props }) => (
   <Container height="auto">
     <Head />
-    <Component {...props} />
+    <FloatFooterPageContentWrapper>
+      <Component {...props} />
+    </FloatFooterPageContentWrapper>
```
hlminh2000: This solution unfortunately would not work out very well as it affects other pages. 
I'd suggest a modification to the cohort builder layout in `src/components/CohortBuilder/index.js` would be a better solution, probably with a `min-height` applied somewhere :) @ciaranschutte is your expert here ;)

screen shot 2019-02-28 at 10 10 44 am

jaouad-benassila: Add code changes to style of  **kf-portal-ui/src/components/CohortBuilder/index.js** to fix this bug

PR #1265, opened 2019-02-28
Reduce the top and bottom padding on the category bar filters

jaouad-benassila:


hlminh2000:
I actually like the how the current content has a lot of room to breath personally :P but code LGTM! :)

PR #1272, opened 2019-03-02
♻️ move overlay to parent component

hlminh2000:
Uses an actual flag for no data rather than faulsey sqon, which semantically means something different. 

Falsey sqon should mean no filter, which means we should show all data, not no data

### DIFF
--- a/src/components/CohortBuilder/ParticipantsTableView/index.js
+++ b/src/components/CohortBuilder/ParticipantsTableView/index.js
@@ -18,14 +18,22 @@ const enhance = compose(
   withState('pageIndex', 'setPageIndex', 0),
 );
 
-const ParticipantsTableView = ({ sqon, api, pageIndex, pageSize, setPageIndex, setPageSize }) => (
+const ParticipantsTableView = ({
+  sqon,
+  api,
+  pageIndex,
+  pageSize,
+  setPageIndex,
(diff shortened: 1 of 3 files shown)


src/components/CohortBuilder/ParticipantsTableView/index.js
```diff
@@ -18,14 +18,22 @@ const enhance = compose(
...
+  noData = false,
+}) => (
   <QueriesResolver api={api} sqon={sqon} queries={[participantsQuery(sqon, pageSize, pageIndex)]}>
     {({ isLoading, data, error }) =>
       error ? (
         <TableErrorView error={error} />
       ) : (
         <React.Fragment>
-          {!sqon ? <EmptyCohortOverlay /> : null}
+          {noData ? <EmptyCohortOverlay /> : null}
```
jberube: Why do we don't just check for `data.length === 0`?

hlminh2000: `[undefined].length` is still 1. 
But you're right, for the table, we can certainly check `data.participant.hits.edges.length`.

There's already a good check from above that works for both the table and the summaries, so wanted to keep the logic centralized. In fact, this overlay should probably go to the parent component, it doesn't need to be here at all

PR #1278, opened 2019-03-04
Fence Updates - DCF and Gen3, User settings and Dashboard

joneubank:
* Converted Gen3 specific services to be fence agnostic, taking a fence repository key as an argument
* Introduced DCF as an integration
* UI changes to accomodate this on the Dashboard and User Settings page


joneubank:
Merging this will require .env file changes. These changes are reflected in teh schema, and I can provide values to devs in the oicr slack.

src/components/Fence/FenceAuthorizedStudies.js
```diff
@@ -105,64 +108,62 @@ const Gen3ProjectList = compose(withApi, withTheme, withHistory)(
...
+      return aggregations ? (
+        projectIds
+          .filter(id =>
+            get(aggregations, `${toGqlString(id)}.participants__study__short_name.buckets.length`),
+          )
+          .map(id => {
+            const studyNameBuckets = get(
+              aggregations,
+              `${toGqlString(id)}.participants__study__short_name.buckets`,
+            );
```
ciaranschutte: Do we need a default value in this `get` as a third param?

src/components/UserProfile/UserIntegrations/FenceIntegrationItem.js
```diff
@@ -0,0 +1,141 @@
...
+
+const enhanceActions = compose(
+  withApi,
+  injectState,
+);
+
+const Actions = enhanceActions(
+  ({ fence, api, effects, loading, setLoading, refresh, connected }) => {
+    return loading ? (
+      <LoadingSpinner heihgt={48} />
```
ciaranschutte: typo: `heihgt -> `height`

src/components/Fence/FenceAuthorizedStudies.js
joneubank: We do, get returns 'undefined' if there is no value, and we immediately assume an array on the next line

Nice catch, I just copied this over from an existing element...

src/components/UserDashboard/AuthorizedStudies/index.js
```diff
@@ -1,112 +1,100 @@
...
-)(({ state: { integrationTokens, loggedInUser }, effects, theme, api, ...props }) => (
-  <Component initialState={{ connected: false, badgeNumber: null }}>
-    {({ setState, state }) => {
-      const Header = <CardHeader title="Authorized Studies" badge={state.badgeNumber} />;
+  lifecycle({
+    componentDidMount() {
+      const { api } = this.props;
+      this.props.effects.fetchFenceConnections({ api });
+    },
+  }),
```
hlminh2000: would it make sense to move this to the `<Component />` below with `didMount`? Just because these big chain of HOCs can get confusing about where things come from, like `this.props.effects.fetchFenceConnections` :P naming `didMount` can help avoid the excessive inline lambdas :)

src/components/FileRepo/FileRepo.js
```diff
@@ -257,7 +258,9 @@ const FileRepo = compose(
 );
 
 export default props => (
-  <Gen3UserProvider
+  //Only has Gen3 integration, needs to be refactored to have all data repository integrations
```
hlminh2000: just wanna confirm we're ok with moving this forward without allowing user downloads :)

src/components/UserProfile/UserIntegrations/IntegrationTableItem.js
```diff
@@ -0,0 +1,70 @@
...
+  td.actions {
+    width: 363px;
+
+    background: ${({ theme }) => theme.backgroundGrey};
+  }
+`;
+
+export default ({ connected = false, logo, description, actions, ...props }) => {
+  return (
+    <Row {...props}>
```
hlminh2000: Not sure what else other than `className` needs to be passed down here, if nothing, I'd suggest going without `...props`. 
Passing unnecessary props down can cause weird errors so just wanna avoid `...props` where possible :)

src/components/FileRepo/FileRepo.js
joneubank: yes this is acceptable. Apparently there wont be direct downloads from the DCF repository, so at this time this is ok.

There is new UI to support showing DCF files with a greyed out download button but this can be tackled in upcoming sprint. For now, file repo is only interested in Gen3 connection.

src/components/UserDashboard/AuthorizedStudies/index.js
joneubank: Can do. No reason to mix usage like I did.

src/components/FileRepo/FileRepo.js
hlminh2000: got it, thanks! :)

src/components/Fence/FenceAuthorizedStudies.js
joneubank: 👍

src/components/UserProfile/UserIntegrations/FenceIntegrationItem.js
joneubank: 👍

src/components/UserProfile/UserIntegrations/IntegrationTableItem.js
joneubank: 👍

src/components/UserDashboard/AuthorizedStudies/index.js
joneubank: 👍

joneubank:
Made a couple structural changes to simplify the state:
*  remove the FenceUserProvider components from the dashboard and make the fence state provided globally
* Update log on component mounts for choosing when to fetch fence user data (don't repeat if already fetched)
* update the file access checking methods to be aware of multiple fences
* make the fence scope dependent on the fence being connected to (DCF is different than Gen3)

src/stateProviders/provideFenceConnections.js
```diff
@@ -0,0 +1,92 @@
...
+  initialState: props => ({
+    fenceConnectionsInitialized: false,
+    fenceConnections: {},
+
+    fenceStudiesInitialized: false,
+    fenceStudies: {},
+  }),
+  computed: {
+    fenceAuthStudies: ({ fenceStudies }) =>
+      !isEmpty(fenceStudies) &&
```
hlminh2000: if this is false, this function will return `false`, is that ok? is it better to return `[]`? it seems we are checking on `fenceAuthStudies.length` anyway around

src/stateProviders/provideFenceConnections.js
```diff
@@ -0,0 +1,92 @@
...
+      flatMap(Object.values(fenceStudies), studies => studies.authorizedStudies),
+    fenceNonAuthStudies: ({ fenceStudies }) =>
+      !isEmpty(fenceStudies) &&
```
hlminh2000: .

src/stateProviders/provideFenceConnections.js
```diff
@@ -0,0 +1,92 @@
...
+            effects.addFenceConnection({ fence, details });
+          })
+          .catch(err => console.log(`Error fetching fence connection for '${fence}': ${err}`)),
+      );
+
+      return Promise.all(fenceConnectionsFetchArray).then(() => {
+        effects.setFenceConnectionsInitialized();
+        effects.setFenceStudiesInitialized(false);
+      });
+    },
```
hlminh2000: would it make sense to handle the fetching in the [initialize](https://github.com/FormidableLabs/freactal#initialize) effect rather than letting a wrapped component trigger this? Same for `fetchFenceStudies`

src/stateProviders/provideFenceConnections.js
hlminh2000: if this is `false`, we'll return `false`, is it better to return `[]`? :)

src/stateProviders/provideFenceConnections.js
hlminh2000: if this is false, we'll return false, is it better to return []? :)

src/stateProviders/provideFenceConnections.js
joneubank: It does make sense, but a couple tricks - it is dependant on the user being logged in so step one is move the provider to a page component that only loads once the user is logged in, and then we need a way to provide the initialized API object that has the JWT auth. If theres a clean way to satisfy those dependencies than I would prefer to run this as initialized, but we'll also need to trigger it on log in as well.

Note that we will still need to trigger the studies load whenever they add a new connection.

PR #1280, opened 2019-03-04
Feature/reduce padding dropdown options

jaouad-benassila:


kcullion:
@jaouad-benassila looks good to me!

PR #1284, opened 2019-03-04
Feature/file breakdown ui fixes

jaouad-benassila:


ciaranschutte:
Header with badge looks off.
You can use the same code in the `Studies Card` it should fix it 

```
    <CohortCard scrollable={true} title={'Available Data'} badge={filesTotal ? filesTotal : null}>
```

ciaranschutte:
There is no margin under the Header anymore.
after
before

ciaranschutte:
File links do not turn a lighter pink when hovered, same as `File Repo`.
Row highlighted grey but anchor doesn't style.

src/components/CohortBuilder/Summary/FileBreakdown.js
```diff
@@ -86,29 +75,46 @@ const generateFileColumnContents = dataset => {
...
+              style: columnStyles,
+            },
+            {
+              Header: 'Files',
+              accessor: 'fileLink',
+              minWidth: 40,
+              style: columnStyles,
+            },
           ]}
+          className="-striped -highlight"
```
ciaranschutte: `-striped` by default so you can just add `-highlight` i think :)

hlminh2000:
> File links do not turn a lighter pink when hovered, same as `File Repo`.
> Row highlighted grey but anchor doesn't style.

regarding this, there's a `Link` component from `uikit/Core` that comes with the styles. Also it wraps react-router's link so we can avoid a page reload :)

src/components/CohortBuilder/Summary/FileBreakdown.js
```diff
@@ -4,24 +4,13 @@ import { withTheme } from 'emotion-theming';
...
-  color: ${({ theme }) => theme.secondary};
-  & a {
-    font-weight: normal;
-    color: ${({ theme }) => theme.hover};
-    text-decoration: underline;
-    margin-left: 5px;
-  }
-`;
+const columnStyles = {
+  margin: '-2% 0',
```
hlminh2000: @ciaranschutte does `BaseDataTable` expose the full `react-table` API? if so, there might be better ways to apply styles to the rows. Regardless, `-2%` margin seems odd to me as it will not be consistent across viewport size, can we try a fixed `px` value? :)

jaouad-benassila: Using % is more responsive than px

hlminh2000: I don't think we want it to respond to the width, it looks odd in a certain screen size, easier to just fix the `px` if possible
screen shot 2019-03-05 at 2 25 35 pm

ciaranschutte: @hlminh2000 not whole API but most.
`style` passed in with `Column` is valid in their API ... I don't like the feel of it either

hlminh2000: ok that's fine then :) 
anyway, `-5px 0` seems to work well for me here

Issue #1076, opened 2019-01-11
Search All Filters


rosibaj:
Zeplin: https://zpl.io/bAZrXyn

![Screen Shot 2019-01-21 at 3.41.49 PM.png](https://images.zenhubusercontent.com/5a5f5bcd4b5806bc2bc31a86/2dba4dba-d428-4d9e-9b3f-7475394c90c2)

PR #1290, opened 2019-03-05
Feature/1221 table selection

jberube:
Cohort Builder Table View selection and actions.
Currently only supports the "remove from cohort".


src/components/CohortBuilder/ParticipantsTableView/index.js
```diff
@@ -14,33 +15,88 @@ const enhance = compose(
...
+      api={api}
+      queries={[participantsQuery(sqon, pageSize, pageIndex)]}
+    >
+      {({ isLoading, data, error }) => {
+        if (error) {
+          return <TableErrorView error={error} />;
+        }
+
+        const isRowSelected = node =>
+          allRowsSelected || selectedRows.some(row => row === node.participantId);
```
hlminh2000: So looks like we are currently not allowing removal of rows from the list once `allRowsSelected` is true. 

I am ok with moving this forward as is regarding this point, but since we have disabled removal of everything from the cohort, it would make sense to allow such use case, else I don't think the use of sets is well justified here. Maybe we would rather allow removal of sets from cohort and allow user to modify the sqon after the fact?

@rosibaj this is something we should discuss, as the use of set is a little more than a convenience factor here.

rosibaj: hmm i think we would want to be able to remove rows from the page once the user selects all.  So for example, i select all, then remove 3. 

im not sure what you mean by `Maybe we would rather allow removal of sets from cohort and allow user to modify the sqon after the fact?`

hlminh2000: Thanks, I agree.
As for the other comment, it's basically enabling the "remove from cohort" button even if everything is selected (results will show 0), then user can still modify the sqon after the fact to get non-0 results. I don't think that's intuitive but it was an option so wanted your thoughts.

src/components/CohortBuilder/index.js
```diff
@@ -50,6 +56,50 @@ const CohortBuilder = () => (
...
+          .then(({ data }) => {
+            const newSqon = {
+              op: 'and',
+              content: [
+                activeSqonIndex,
+                {
+                  op: 'not',
+                  content: [
+                    {
+                      op: 'in',
```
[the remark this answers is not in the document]
jberube: I remember having some weird results with `not-in`, though it might be the sqon using the `set`s that was faulty at that time.
I'll give it another shot.

src/components/CohortBuilder/Results.js
```diff
@@ -136,6 +145,12 @@ const Results = ({ activeView, activeSqonIndex, setActiveView, theme, sqon, api
   </QueriesResolver>
 );
 
+QueriesResolver.propTypes = {
```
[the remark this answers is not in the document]
jberube: Copy-paste error :man_facepalming:
I'll fix that.

src/components/CohortBuilder/ParticipantsTableView/index.js
jberube: Makes sense. I'll fix the comment on using `not-in` first and see if I can fix that one afterward.

rosibaj: @hlminh2000 so in this scenario `then user can still modify the sqon after the fact to get non-0 results` would the user see the regular dropdown list of all the participant ids and then be able to select some ids?

src/components/CohortBuilder/index.js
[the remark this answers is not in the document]
hlminh2000: If weird results are coming back, it's likely a back-end issue. We can move forward with `not-in` for the UI and investigate the back-end in a separate task :)

src/components/CohortBuilder/ParticipantsTableView/index.js
hlminh2000: no, but that's a great point, the filter component doesn't know how to handle sets

jberube:
I'll make more progress in #1076 before coming back to this one to avoid loosing focus.

src/components/CohortBuilder/ParticipantsTableView/index.js
jberube: After changes to the way the sets and the sqon is built after removing participants, combining removed participants with other queries works nicely.
The way that the sets are represented in the cohort builder is still not super user-friendly, then.

rosibaj: @jberube @hlminh2000 I would vote move forward so we can interact with this in the cohort builder and then we can adjust behaviour - is that possible?

jberube: @hlminh2000 if you agree with @rosibaj , you know which button to push ;)

src/components/CohortBuilder/index.js
```diff
@@ -50,6 +56,45 @@ const CohortBuilder = () => (
       const categoriesSqonUpdate = newSqon => {
         mergeSqonToActiveIndex(newSqon);
       };
+
+      const createNewSqonExcludingParticipants = participantIds => {
+        saveSet({
+          type: 'participant',
+          sqon: {
+            op: 'not-in',
```
hlminh2000: this should be `not-in`

src/components/CohortBuilder/index.js
```diff
@@ -50,6 +56,45 @@ const CohortBuilder = () => (
...
+          path: 'kf_id',
+          api: graphql(api),
+        })
+          .then(({ data }) => {
+            const newSqon = {
+              op: 'and',
+              content: [
+                activeSqonIndex,
+                {
+                  op: 'in',
```
hlminh2000: this should be `not-in`

jberube: I don't see why both the set *and* the query including the set should use `not-in`, it sounds illogical to me.

hlminh2000: oh oops, bad comment again... corrected!

PR #1295, opened 2019-03-05
Remove the Phenotypes Template card

jaouad-benassila:


jaouad-benassila:
I don't know if there is need for this card in the future or not ?

ciaranschutte:
@jaouad-benassila We just need to make the cards remaining take up the whole width of the container by removing the extra dom el and increasing the Col xs space

jaouad-benassila:
remove

hlminh2000:
haha @kcullion I feel like we might want to make the cards taller with all the new real estate eh :)

kcullion:
@hlminh2000 hmmm - I would say just a bit taller so you can see the full "p" of the word Participants on the x-axis. Let's see how that looks first.
screen shot 2019-03-06 at 9 28 41 am

PR #1296, opened 2019-03-05
removes unneeded wrapper

hlminh2000:
Tiny PR :)


ciaranschutte:
May as well move the margins back to the wrapper instead of passing it through styles?

PR #1298, opened 2019-03-06
QueriesResolver race condition

hlminh2000:
2 things here:
1) implement caching with a global cache store so it persists throughout application lifecycle (a highly simplified solution)
2) QueriesResolver updates were allowed to happen in parallel so there was room for race conditions. Fixed this with a task queue.


ciaranschutte:
@@@#---> Double Approval <---#@@@

PR #1306, opened 2019-03-06
Fix ui summary section for 0 participant

jaouad-benassila:


hlminh2000:
@jaouad-benassila does my suggestion make sense? :)

jaouad-benassila:
@hlminh2000 I tried to remove z-index from EmptyCohortOverlay and move its rendering, but it does not work for me

hlminh2000:
@jaouad-benassila a "no z-index" proposal for the problem for this PR :) https://github.com/kids-first/kf-portal-ui/pull/1316
If all looks good, I prefer that over changing z-index as it can become hard to keep up with the z-index race over time :)

PR #1311, opened 2019-03-08
Bug/#1281 histogram colors

ciaranschutte:


src/components/CohortBuilder/Summary/StudiesChart.js
```diff
@@ -100,7 +102,7 @@ const StudiesChart = ({ studies, sqon, theme, api }) => (
               tooltipFormatter={studiesToolTip}
               sortBy={sortDescParticipant}
               tickInterval={4}
-              colors={[theme.chartColors.blue, theme.chartColors.purple]}
+              colors={getCohortBarColors(data)}
```
hlminh2000: I'd suggest feeding `theme` to `getCohortBarColors` and reference it in there rather than importing `chartColors ` directly in `src/components/CohortBuilder/Summary/ui.js`. For consistency with the usage of theme :)

ciaranschutte: @hlminh2000 updated :)

PR #1316, opened 2019-03-11
♻️ Resolves the layering issue without modifying z-index

hlminh2000:


jaouad-benassila:
It works fine, I prefer this fix  than to use z-index.

PR #1321, opened 2019-03-12
♻️ prefers variable over interpolation for dynamic gql values

hlminh2000:


hlminh2000:
@ciaranschutte approves!

PR #1331, opened 2019-03-12
✨ saving and loading virtual studies

hlminh2000:
Needs kf-persona to be deployed first.

Another consideration worth noting: riff currently can only handle one type of entity. 
The dashboard is currently pulling everything associated with a given user, so a quick hack was needed there to filter out irrelevant entries. 

Should Riff be updated to handle multiple entities? Maybe with a tag field? @joneubank 
Point of interest: https://github.com/kids-first/kf-portal-ui/pull/1331/files#diff-e8f8d55bc002d717a912d810a546e4f0R81


src/components/CohortBuilder/SQONProvider.js
```diff
@@ -48,12 +89,31 @@ const SQONProvider = ({ children }) => {
...
+      ...history.location,
+      search: stringify({ id }),
+    });
+    syncWithHistory(s);
+  };
 
   return (
-    <Component initialState={initialState} didMount={didMount} didUpdate={didUpdate}>
+    <Component
+      virtualStudyId={virtualStudyId}
```
ciaranschutte: `virtualStudyId` need to be passed into react-component-component?

hlminh2000: so `didUpdate` receives it as props :)

PR #1334, opened 2019-03-13
Feature/#1307 loaders

ciaranschutte:
@hlminh2000 @jberube I'm trying to figure out a cleaner way to do the loading states other than having multiple conditions to render a `<LoaderCard/>` component.

We just want a spinner in the card.
So data passed down from Summary may be null so we need to pass it's loading state into each card.
Then each card has it's `QueryResolver` so it will also have a local loading state

Things get even messier in `FileBreakdown` where we are nesting multiple `QueryResolvers`

Any thoughts?


hlminh2000:
Some ideas:
1) We can implement the loading render as a prop built into the `Card` itself, defaulting to `false` then we can do:
```
<Car loading={loadingThis && loadingThat}/>
```
This has the benefit of enforcing loading state style consistency too.
2) For nested loading state, we can pass the parent's `loading` down to the child as prop, which renames it to `parentLoading`, and pass down its internal `loading` state down as `loading && parentLoading` at each level, so ultimately when it reaches the card, it looks like `Card loading={loading && parentLoading}`

src/components/CohortBuilder/Summary/DiagnosesChart.js
```diff
@@ -74,42 +74,45 @@ const toSingleDiagQueries = ({ topDiagnoses, sqon }) =>
...
+  isLoadingSummary ? (
+    <LoaderCard title="Most Frequent Diagnoses" />
+  ) : (
+    <QueriesResolver
+      name="GQL_DIAGNOSIS_CHART"
+      api={api}
+      queries={toSingleDiagQueries({ topDiagnoses, sqon })}
+    >
+      {({ isLoading, data }) =>
+        isLoading ? (
```
hlminh2000: can we do `isLoadingSummary || isLoading` here instead of a layer above?

jberube:
Generally speaking, I think this is upside down : it's not the parent that should decide to either render a Loader or a Card - or an error, or a "no data" state for that matter - but the Card that should decide to render one of it's 4 states: a loader, a "no data" state, an error or it's normal content.
If some specialization of a Card share the way they display the 3 special states, then you can wrap those implementations of a Card in a e.g. <CardWrapper> that receives data, loading, error and decide what to display. That <CardWrapper> should receive the <Card> as a children since it's it's normal state, but render the loader, error or "no data" insteadl, when necessary.

PR #1336, opened 2019-03-15
Remove upload id

rosibaj:
Commenting out the upload id section as it will not be completed for AACR release


rosibaj:
forgot to select correct merge branch - closing

PR #1338, opened 2019-03-18
Feature/download button table view

jaouad-benassila:


src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
```diff
@@ -163,11 +181,24 @@ class ParticipantsTable extends Component {
     } = this.props;
     const { columns } = this.state;
     const selectedRowsCount = allRowsSelected ? dataTotalCount : selectedRows.length;
+    const projectId = arrangerProjectId;
 
     const handleRemoveFromCohort = () => {
       onRemoveFromCohort();
     };
 
+    const selectionSQON = this.props.selectedRows.length
```
jberube: You can safely remove `replaceSQON` here since it does nothing.

```
export const replaceSQON: TMergeSQON = (q, ctxq) => {
  if (!ctxq && !q) return null;
  if (!ctxq) return q;
  if (!q) return ctxq;
```

src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
```diff
@@ -196,7 +196,7 @@ class ParticipantsTable extends Component {
             },
           ],
         }
-      : undefined;
+      : this.props.sqon;
```
hlminh2000: I don't think `sqon` is passed in from outside yet

PR #1346, opened 2019-03-18
order the available data card alphabetically by data type

jaouad-benassila:


src/components/CohortBuilder/Summary/FileBreakdown/index.js
```diff
@@ -114,9 +114,18 @@ export const fileBreakdownQuery = sqon => ({
...
+  const dataTypeB = b.dataType.toUpperCase();
+  if (dataTypeA > dataTypeB) return 1;
+  if (dataTypeB > dataTypeA) return -1;
+  return 0;
+};
+
 const FileBreakdown = ({ fileDataTypes, sqon, theme, state, api }) => (
   <FileBreakdownQueryResolver sqon={sqon} fileDataTypes={fileDataTypes}>
     {({ data, isLoading }) => {
+      data = data.sort(compareElement);
```
hlminh2000: no need for assignment here as `data.sort` does an in-place sort :) 
Also we would actually prefer lodash's `sortBy` to avoid mutation, something like `sortBy(data, ({dataType}) => dataType.toUpperCase())` should do the job, I think :P

jaouad-benassila: Ok, I will use lodash to sort array

PR #1348, opened 2019-03-19
✨ sharing virtual studies

hlminh2000:


src/components/ShareSaveQuery/SaveQuery.js
```diff
@@ -73,163 +78,162 @@ export default injectState(
...
-                      <Component
-                        initialState={{ saved: false }}
-                        render={({ state, setState }) => (
+          <WhiteButton
+            disabled={disabled}
+            onClick={() =>
+              !disabled &&
+              this.setState({ open: true }, () => {
+                // so hacky, but couldn't get any reasonable approach to work
+                let id = setInterval(() => {
```
jberube: At least, add a timestamp before, and stop if the interval is running since more than enough time, to avoid memory leaks.

hlminh2000: great point :) although this was actually unmodified existing code, I just removed the wrapping `<ButtonContainer/>` to support layout :P

src/components/ShareSaveQuery/ShareQuery.js
```diff
@@ -48,136 +47,136 @@ export default injectState(
...
-      shortenApi({ stats, sqon, loggedInUser, api, sharedPublicly: true })
-        .then(data => {
-          this.setState({
-            link: urlJoin(shortUrlResolveRoot, data.id),
-          });
-        })
-        .catch(error => {
-          this.setState({ error: true });
+    share = async () => {
+      let {
```
ciaranschutte: why `let`?

hlminh2000: existing code :P frankly, not sure

PR #1350, opened 2019-03-19
Feature/#1307 loading loaders

ciaranschutte:
Loading States for cards.

Show loader if any parent loading a request.

Changed QueriesResolver to not make an empty query
i.e one with [] as request body, this is an invalid POST body
This solves the issue of resolvers making the queries with their default [] state

### DIFF
--- a/src/components/CohortBuilder/QueriesResolver.js
+++ b/src/components/CohortBuilder/QueriesResolver.js
@@ -34,7 +34,7 @@ class QueriesResolver extends React.Component {
 
   update = () => {
     const { queries = [], useCache = true } = this.props;
-    queries.length !== 0
+    return queries.length !== 0
       ? this.taskQueue.enqueue(
           () =>
             new Promise(async resolve => {
--- a/src/components/CohortBuilder/QueriesResolver.js
+++ b/src/components/CohortBuilder/QueriesResolver.js
@@ -32,29 +32,33 @@ class QueriesResolver extends React.Component {
     }
   }
 
-  update = () =>
-    this.taskQueue.enqueue(
-      () =>
-        new Promise(async resolve => {
-          const { queries = [], useCache = true } = this.props;
-          if (queries.length === 0) return resolve();
-          const body = JSON.stringify(
-            queries.map(q => ({
-              query: typeof q.query === 'string' ? q.query : print(q.query),
-              variables: q.variables,
-            })),
-          );
-          try {
-            if (!useCache) {
-              this.setState({ isLoading: true });
-            }
-            const data = useCache ? await this.cachedFetchData(body) : await this.fetchData(body);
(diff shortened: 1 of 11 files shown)


src/components/CohortBuilder/QueriesResolver.js
```diff
@@ -32,28 +32,33 @@ class QueriesResolver extends React.Component {
...
+          () =>
+            new Promise(async resolve => {
+              const body = JSON.stringify(
+                queries.map(q => ({
+                  query: typeof q.query === 'string' ? q.query : print(q.query),
+                  variables: q.variables,
+                })),
+              );
+              try {
+                if (!useCache) {
```
hlminh2000: This was my hack, but `useCache` shouldn't be used here really, let's use its own `useLoadingState` prop based on our earlier chat?

src/components/CohortBuilder/Summary/index.js
```diff
@@ -44,38 +43,45 @@ const Summary = ({
...
-                <DemographicChart data={demographicData} />
+                <DemographicChart data={demographicData} isParentLoading={isLoading} />
               </PaddedColumn>
               <PaddedColumn md={spacing.md} lg={spacing.lg}>
-                <SurvivalChart data={survivalPlotMock} />
+                <SurvivalChart data={survivalPlotMock} isParentLoading={isLoading} />
               </PaddedColumn>{' '}
               <PaddedColumn md={spacing.md} lg={spacing.lg}>
-                <AgeDiagChart data={ageDiagData} />
+                <AgeDiagChart data={ageDiagData} isParentLoading={isLoading} />
```
hlminh2000: Does it make more sense to expose `isParentLoading` as just `isLoading` instead? Totally just a naming thing, but I think it reads better :)
Each card can rename it internally to account for namespacing

ciaranschutte: Sure!

src/components/CohortBuilder/QueriesResolver.js
ciaranschutte: @hlminh2000 
So I ended up only needing to check for parent loading state for the Cards and not having to add a new loading state to QueriesResolver, so useCache is just the same as before. Only change to QueriesResolver is the no query on empty array. Do you want me to just take out the useCache setstate?

jberube:
```
Changed QueriesResolver to not make an empty query
i.e one with [] as request body, this is an invalid POST body
This solves the issue of resolvers making the queries with their default [] state
```
:+1:

src/components/CohortBuilder/Summary/index.js
```diff
@@ -44,38 +43,45 @@ const Summary = ({
     ]}
   >
     {({ isLoading, data }) => {
```
jberube: if `data` can be `null`, it should either have a default here or on line 56.

PR #1352, opened 2019-03-19
add option to not replace text with loader, display to side instead

ciaranschutte:
Adds option to display loader to side of link
Looks better in Results header instead of `6,432 Participants with <Loader>`,
now it is `6,432 Participants with 2,232 Files <Loader>`


jberube:
LGTM, but has a merge conflict to be dealt with.

PR #1372, opened 2019-03-20
🐛 filters out riff entries that are not saved queries

hlminh2000:
stop gap solution to allow dashboard SavedQueries card to identify saved queries from riff data.

### DIFF
--- a/src/components/UserDashboard/SavedQueries/index.js
+++ b/src/components/UserDashboard/SavedQueries/index.js
@@ -28,6 +28,12 @@ const FileRepositoryLink = styled(Link)`
   color: ${({ theme }) => theme.primary};
 `;
 
+/**
+ * This is a stop-gap solution until Riff supports tags to support
+ * server-side differentiation between entity types
+ */
+const isSavedQuery = q => !!q.content.Files;
+
 export const MySavedQueries = compose(
   provideSavedQueries,
   injectState,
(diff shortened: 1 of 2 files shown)


src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -28,6 +28,12 @@ const FileRepositoryLink = styled(Link)`
   color: ${({ theme }) => theme.primary};
 `;
 
+/**
```
joneubank: 👍 Thank you for comments like these.

src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -44,7 +50,8 @@ export const MySavedQueries = compose(
     api,
     theme,
   }) => {
-    const Header = <CardHeader title="Saved Queries" badge={queries.length} />;
+    const savedQueries = queries.filter(isSavedQuery);
```
joneubank: Should this filtering be done in the service that fetches the queries? The component doesn't want to own the logic of separating queries from v-studies

joneubank: Just looked at the state provider, it doesnt separate the fetching into a Riff service file. So can just move the filtering there.......

hlminh2000: great call! :) thanks!

PR #1435, opened 2019-03-25
remove default button styling

ciaranschutte:
I'd appreciate it if someone not on Linux could check this is looking correct please :)


src/components/CohortBuilder/SqonBuilder/SqonBuilder.css
```diff
@@ -0,0 +1,7 @@
+.sqonBuilder .sqonListActionButton {
+  background-color: transparent;
+}
+
+.sqonBuilder .actionHeaderContainer button {
+  background-color: transparent;
+}
```
hlminh2000: let's do this in the `SqonBuilderContainer` styled component in `src/components/CohortBuilder/SqonBuilder/index.js` instead of a css file :)

PR #1437, opened 2019-03-26
add showCursor false to vertical and horizontal bars

ciaranschutte:


src/chartkit/components/VerticalBar.js
```diff
@@ -37,7 +37,10 @@ class VerticalBar extends Component {
   }
 
   onMouseEnter(data, e) {
-    e.target.style.cursor = 'pointer';
+    const { showCursor = true } = this.props;
+    if (showCursor) {
+      e.target.style.cursor = 'pointer';
+    }
```
hlminh2000: I would also clean up this side effect in `onMouseLeave`. The more idiomatic React approach to this would be to handle this in `componentDidMount` and `componentDidUpdate`, and cover the `false` case as well.

jberube: It look to me like something that should be handled by css only, using `:hover`

hlminh2000: yeah totally agreed if the elements are accessible by `css`. I believe they're svg elements though so figured @ciaranschutte would know best :)

ciaranschutte: yeah svg in nivo chart so CSS hover won't work

hlminh2000: please ignore my comment, we do not have access to the element outside of nivo's callback. Apologies for the confusion :)

PR #1449, opened 2019-03-26
Fix/1359 Fixed the "select all" of participant table

jberube:
This is by no mean a clean fix, but I have to work around react-table bugs and limitations.

To be done in a better way once react-table is updated.


ciaranschutte:
@jberube Can you change the branch to `feature/cohort-builder` please? :)

jberube:
Thanks @ciaranschutte for pointing that out!

PR #1486, opened 2019-03-28
Fix/1405 row colors

jberube:
Only did the alternate colors to match the File Repo table.

Neither have a color for selected rows.


ciaranschutte:
@jberube change base branch please :) I don't think I can do without being author of pr :'(

PR #1501, opened 2019-03-29
pulls studies from global state

hlminh2000:


joneubank:
That was a dope refactor with the fenceConnection initializer HOC

PR #1523, opened 2019-03-30
fixes the grouping of authorized vs unauthorized studies and files

hlminh2000:


src/components/UserDashboard/AuthorizedStudies/StudiesConnected.js
```diff
@@ -16,7 +16,7 @@ import { createStudyIdSqon, createAcceptedFilesByUserStudySqon } from 'services/
 import Study from './Study';
 import { trackUserInteraction, TRACKING_EVENTS } from 'services/analyticsTracking';
 
-import { isEmpty } from 'lodash';
+import _, { isEmpty } from 'lodash';
```
joneubank: might as well just import `get, keys` directly?

src/services/fileAccessControl.js
```diff
@@ -1,4 +1,4 @@
-import { get, isObject } from 'lodash';
+import _, { get, isObject } from 'lodash';
```
joneubank: Importing all of lodash is no issue, just would rather be consistent. In large files,`_.whatever` is a useful syntax to know where the method comes from. In either case, consistency within a file at least

Issue #1438, opened 2019-03-26
Facets: When sending in an empty array, disable button


rosibaj:
The user should not be able to apply a continuous filters if they haven't filled out the input fields. Currently if I leave them blank, they are applied to the query area and actually filter the results.
![Screen Shot 2019-03-18 at 10 41 56 AM](https://user-images.githubusercontent.com/18424138/54538464-ac71a200-496a-11e9-8b22-d0b43490a407.png)

**EXPECTED BEHAVIOUR:**

- [ ] When sending in an empty array (i.e the user has not entered any values), disable the `Apply` button so they cannot add an empty pill
![image](https://user-images.githubusercontent.com/15824252/55337802-570aba00-546d-11e9-9299-1856a5784018.png)

You can find the disabled "APPLY" button in this mock: https://zpl.io/2jQj8Pp
- [ ] Add a tooltip that says `Please select at least one value to apply`

vferretti:
I would rather disable the button "Apply" if no value is entered.

rosibaj:
@vferretti Yes, that is what the ticket is describing! I will make it more clear.

nyanofthemoon:
branch : bug/disable-apply-button-on-empty-query

nyanofthemoon:
@aalex @jaouad-benassila @jberube https://github.com/kids-first/kf-portal-ui/pull/1529

nyanofthemoon:
@aalex deployed to https://kf-qa-portal.netlify.com/

aalex:
It works for me for the checkboxes. Good work!

For selecting a range, though, it seems like the "Apply" button never becomes clickable. Am I doing something wrong?

![screenshot_2019-04-03_15-35-57-cannot-click-apply-anymore.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/c907cc62-783e-4627-8d23-a85c112755c4)

aalex:
Should also fix #1532 when merged.

aalex:
I've tried it in the kf-qa-portal environment, and I need to click outside of the number box (text field) in order to enable the apply button. Would it be possible to avoid having to do that?

![screenshot_2019-04-05_13-34-45_click_outside_numberbox.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/c626db28-cfd3-4e9a-92ea-ac79e136884d)

(I've also posted this comment to another ticket, but they were duplicates, so I think it rather goes here)

[...]

PR #1529, opened 2019-04-01
Facets: When sending in an empty array, disable button #1438

nyanofthemoon:


nyanofthemoon:
Same as https://github.com/kids-first/kf-portal-ui/pull/1529?

jberube:
> Same as #1529?

This IS #1529

PR #1538, opened 2019-04-02
:sparkles: Cavatica action per file repo row

joneubank:
This also makes a Component that handles opening the Cavatica modal that is separate from the UI of the button.

### DIFF
--- a/src/components/FileRepo/CustomColumns/ActionsColumn.js
+++ b/src/components/FileRepo/CustomColumns/ActionsColumn.js
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import Component from 'react-component-component';
+import styled from 'react-emotion';
 import { get, intersection } from 'lodash';
 import { compose } from 'recompose';
 import { withTheme } from 'emotion-theming';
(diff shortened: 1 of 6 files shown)


src/components/FileRepo/CustomColumns/ActionsColumn.js
```diff
@@ -68,7 +76,14 @@ const ActionItems = ({ value, fence, hasAccess, theme }) => {
             <ControlledIcon fill={theme.lightBlue} />
           </Tooltip>
         )}
-      </Column>
+      </ButtonWrapper>
+      <ButtonWrapper>
+        {hasAccess && (
+          <CavaticaOpenModalWrapper fileIds={[value]}>
+            <CavaticaLogo fill={theme.lightBlue} width={16} />
+          </CavaticaOpenModalWrapper>
```
hlminh2000: Do we wanna show something else if `!hasAccess`? Or maybe should we just move the whole `hasAccess` check wrap both the fence thing and cavatica?

joneubank: The design in this case just has nothing. Ask Kim?

PR #1545, opened 2019-04-03
wrap text inside cells

jaouad-benassila:


jberube:
Please link to the issue

PR #1550, opened 2019-04-04
make search all filter bigger

jaouad-benassila:


jberube:
Please link to the issue

PR #1553, opened 2019-04-05
Add color to selected row of participant table

jaouad-benassila:


jberube:
Please link to the issue

jberube:
Could you rather put a class on the selected rows and set the color with CSS instead so the component is reusable, or the color can be overiden?

PR #1554, opened 2019-04-08
fix style  button analyze file in cavatica

jaouad-benassila:


jberube:
Please link to the issue

PR #1560, opened 2019-04-09
Integrate with new Arranger admin system

hlminh2000:
Integrating the new arranger admin UI to replace the legacy one.

This is effectively a separate app from the existing platform, so trying to keep it relatively decoupled from the rest of the `kf-portal-ui`, in case future needs arises to have this deployed separately altogether.
- Lazy loading to avoid bloating up the bundle size, also a good example for implementing code split in future.
- Leaving the legacy admin dashboard around for now, will be removed once all finalized

### DIFF
--- a/src/common/injectGlobals.js
+++ b/src/common/injectGlobals.js
@@ -91,3 +91,5 @@ export const reactApiDataVersionFallback: string =
 
 // Public Stats
 export const publicStatsApiRoot = getApplicationEnvVar('PUBLIC_STATS_ROOT') || '';
+
+export const arrangerAdminApiRoot = getApplicationEnvVar('ARRANGER_ADMIN_ROOT') || null;
--- a/src/components/ArrangerAdmin.js
+++ b/src/components/ArrangerAdmin.js
@@ -1,0 +1,30 @@
+import React, {Suspense} from 'react';
+import { arrangerAdminApiRoot } from 'common/injectGlobals';
+import { EGO_JWT_KEY } from 'common/constants';
+import { withRouter } from 'react-router-dom';
+
+const LazyArrangerAdminUi = React.lazy(() => import('@arranger/admin-ui/dist'))
+
+export default withRouter(({ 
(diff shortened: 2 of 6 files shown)


src/components/ArrangerAdmin.js
```diff
@@ -0,0 +1,30 @@
...
+    <Suspense fallback={<div>Loading...</div>}>
+      <LazyArrangerAdminUi 
+        basename={baseRoute} 
+        apiRoot={arrangerAdminApiRoot} 
+        fetcher={(url, config) => 
+          fetch(url, {
+            ...config,
+            headers: {
+              ...config.headers,
+              authorization: `Bearer ${localStorage[EGO_JWT_KEY]}`
```
jberube: Can you use the standard `.getItem(key)` signature, please?

src/components/ArrangerAdmin.js
```diff
@@ -0,0 +1,30 @@
+import React, {Suspense} from 'react';
+import { arrangerAdminApiRoot } from 'common/injectGlobals';
+import { EGO_JWT_KEY } from 'common/constants';
+import { withRouter } from 'react-router-dom';
+
+const LazyArrangerAdminUi = React.lazy(() => import('@arranger/admin-ui/dist'))
```
jberube: Didn't knew that worked with Webpack 3!

src/common/injectGlobals.js
```diff
@@ -91,3 +91,5 @@ export const reactApiDataVersionFallback: string =
 
 // Public Stats
 export const publicStatsApiRoot = getApplicationEnvVar('PUBLIC_STATS_ROOT') || '';
+
+export const arrangerAdminApiRoot = getApplicationEnvVar('ARRANGER_ADMIN_ROOT') || null;
```
jberube: Please don't forget to send us the value for `ARRANGER_ADMIN_ROOT` ;)

PR #1567, opened 2019-04-10
fix total participant table export

jaouad-benassila:


jaouad-benassila:
Did you test the fix ? it's not problem of back-end
1 -  it works as expected and the server send all data first time. 
2-  tsv file display all participants.
3-  I call for all data just first time because QueriesResolver has a cache

jberube:
Have you tried it with a cohort with a lot (like all) of the participant?

jaouad-benassila:
yes with 4,005 Participants and it works fast

src/components/CohortBuilder/ParticipantsTableView/index.js
```diff
@@ -54,68 +54,91 @@ const ParticipantsTableView = ({
           );
         }
 
+        const dataTotalCount = data[0] ? data[0].total : 0;
```
nyanofthemoon: @jaouad-benassila I just read the ticket- it says to export "per page" rather than the entire dataset.

src/components/CohortBuilder/ParticipantsTableView/index.js
```diff
@@ -54,68 +54,91 @@ const ParticipantsTableView = ({
...
-                onRemoveFromCohort(selectedRows);
-                // clear selection
-                setAllRowsSelected(s => false);
-                setSelectedRows(s => []);
-              }}
-              selectedRows={selectedRows}
-              allRowsSelected={allRowsSelected}
-            />
-          </Card>
+          <QueriesResolver
```
nyanofthemoon: @jaouad-benassila Could we re-use the already "downloaded" JSON dataset for the current page?

nyanofthemoon:
> 3- I call for all data just first time because QueriesResolver has a cache

Ah, yes, that it does.
...but everytime we change the query we will get fetch entire dataset- imagine when a user will query all of the participants, that might implode the browser unto itself.

jaouad-benassila:
@nyanofthemoon Vincent need total export not partial
we can't re-use the same downloaded json because it's not the same context.
with  QueriesResolver and his cache we get fetch all dataset one time only.

jaouad-benassila:
tested with 8337 Participants and it works as expected

Issue #1575, opened 2019-04-11
Show only one item when they are all the same


aalex:
There are cases where a file has multiple types or a participant has multiple diagnoses. This creates problems in the way we display rows in both the file and participant tables respectively.

## Steps to reproduce
In the participant table, some fields have the same value more than once, for example "Cancer Cancer Cancer" - maybe the participant has been diagnosed three times with the same disease.

Instead, we would like to see that value only once: "Cancer".

If not all values are different, no change is required. We will still show them all.

## Acceptance criterias
- [ ] if this value is unique, display it only once

## See also
- #1570

jberube:
This was done this way to ensure that data related to the same diagnosis are aligned.
E.g. One participant can have 2 diagnoses of the same diagnosis category, and then a third one of a different category.
```
Cat1   Diag1   <- related
Cat1   Diag2   <- related
Cat2   Diag3   <- related
```
That fix creates that issue:

```
Cat1   Diag1   <- related
Cat2   Diag2   <- unrelated
----   Diag3   <- orphan 
```

aalex:
If two are the same, but only one is different, show them all. (important)

Show only one item **only if** all of them are the same. (and show only one item in the first position)

aalex:
Is this still blocked?

nyanofthemoon:
@jaouad-benassila Ok for your pull request https://github.com/kids-first/kf-portal-ui/pull/1577 - please just fix the merge conflicts before you squash and merge!

PR #1577, opened 2019-04-12
fix participant table ui and add tooltip

jaouad-benassila:
this PR is connected to this ticket too:
Show only one item when they are all the same #1575

### DIFF
--- a/src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
+++ b/src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
@@ -22,8 +22,6 @@ import { configureCols } from 'uikit/DataTable/utils/columns';
 import DownloadButton from 'components/FileRepo/DownloadButton';
 import { arrangerProjectId } from 'common/injectGlobals';
 import { SORTABLE_FIELDS_MAPPING } from './queries';
-import { Tooltip } from 'react-tippy';
-import 'react-tippy/dist/tippy.css';
 import { union } from 'lodash';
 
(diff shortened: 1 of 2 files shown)


src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
```diff
@@ -43,27 +46,35 @@ const SelectionCell = ({ value: checked, onCellSelected, row }) => {
 
 const enhance = compose(withState('collapsed', 'setCollapsed', true));
 const CollapsibleMultiLineCell = enhance(({ value: data, collapsed, setCollapsed }) => {
-  // Display a fourth row when there is exactly 4 rows.
+  // Display one row when there is exactly more than one row.
   // Collapsing a single don't save any space.
-  const displayedRowCount = collapsed ? (data.length === 4 ? 4 : 3) : data.length;
-  const displayMoreButton = data.length > 4;
+  const sortedData = union(data);
```
aalex: We will have to remove this call to union, since it doesn't meet our requirements. Speaking of requirements, we will move the removal of duplicates into another story.

aalex:
Some of there changes are for https://github.com/kids-first/kf-portal-ui/issues/1575 - which is about showing only one item if all items are the same. As soon as two or more items are different, we must show them all, though.

src/components/CohortBuilder/ParticipantsTableView/ParticipantsTable.js
nyanofthemoon: @aalex said it's fine now :)

PR #1596, opened 2019-04-16
Charts now available!

joneubank:
Removes placeholder card over survival chart.


jberube:
Never mind my previous comment, I was missing part of the conversation.

Issue #1620, opened 2019-04-18
🐛 Fix hotjar js initialization 


bdolly:
fix the async initialization of the hotjar script loading on page load. Currently it is throwing the below console error 
<img width="381" alt="Screen Shot 2019-04-18 at 1 57 51 PM" src="https://user-images.githubusercontent.com/1334131/56381252-58b0dd80-61e2-11e9-9396-f0fc339c54dc.png">

Need to reconcile this error and ensure the proper loading of the script so that the hotjars recordings and feedback features will be enabled for the portal

PR #1622, opened 2019-04-18
🚑Hotfix/hotjar init error

bdolly:
## Motivation 
The hotjar script loading was throwing an error in the console and preventing any of the hotjars analytics tracking features from being enabled. We want to fix this so that we can start recording user sessions and enable the feedback widget. 

## Implementation details 
Moves the script loading from the analyticsTracking service into the document head using the standard [hotjars script tag](https://help.hotjar.com/hc/en-us/articles/115009336727-How-to-Install-your-Hotjar-Tracking-Code)
deprecates the need for the `REACT_APP_HOTJAR_HOST`


jberube:
That would be related to https://github.com/kids-first/kf-portal-ui/issues/1620 right?

bdolly:
> Could the script that load the hj script be at the end of the body rather than in the head?

It could be at the end but it's set to load `async` so it's non-blocking in the head

bdolly:
> That would be related to #1620 right?

Yup this addresses that issue and would allow us to add the Hotjar feedback widget to the portal

jberube:
Ok, LGTM, I'll wait until after we push the "white page" hotfix to production before merging, though, since we have no more time to test and Vincent has a demo coming.

jberube:
Just realized that this PR points to `master` when it should be pointing to `next`.

I'll try to fix it real quick.

bdolly:
@jberube change based to `next`, would you mind checking out the conflict with `src/components/CohortBuilder/DeleteVirtualStudiesModalContent.js` as I'm not familiar with that file

jberube:
Closed. Has been replaced by #1669 .

PR #1638, opened 2019-04-23
update DCF logo

jaouad-benassila:


aalex:
LGTM. The logo actually looks fine in the browser, even though its blue circle background is missing when I open it with Inkscape.

PR #1649, opened 2019-04-25
Feature/ego upgrade

steptek:


src/components/Login.js
```diff
@@ -31,16 +31,22 @@ import { createExampleQueries } from 'services/riffQueries';
 
 export const isAdminToken = ({ validatedPayload }) => {
   if (!validatedPayload) return false;
-  return get(validatedPayload, 'context.user.roles', []).includes('ADMIN');
+  const jwtUser = get(validatedPayload, 'context.user', {});
+  const roles = jwtUser.roles;
+  const type = jwtUser.type;
+  // maintain backward compatibility
+  return roles && null !== roles && isArrayLikeObject(roles)
```
nyanofthemoon: @steptek `roles && null !== roles &&` might not be required since `isArrayLikeObject` is supposed to return `true` or `false`?

steptek: a double checking..

PR #1659, opened 2019-04-30
Feature/export some rows participant table

jaouad-benassila:


nyanofthemoon:
@jaouad-benassila GTG 🦄

PR #1669, opened 2019-05-02
🚑 fixes the loading of the hotjar js script

jberube:
Removes the fancy scriptjs loading in favor of a plain ol' script tag in the document head.

This is a second PR related to https://github.com/kids-first/kf-portal-ui/pull/1622 to address issuess we had after rebasing the PR.

All credits goes to Ben Dolly :)


jberube:
Tested by @bdolly

Issue #1533, opened 2019-04-02
Save a virtual study: both save and save as


aalex:
As a user, I want to be able to do save a virtual study as a new one.

## Acceptance criterias
- [x] Each user can only save their own studies. (that will need to be done in the backend using the UUID of the user)
- [x] Disable buttons when it's not our's: delete, share, save. ("save as" is possible, though) Check the ownership in the frontend. (it will also be checked in the backend)
- [x] "Save as" must be as the right of "Save"
- [x] Use the fa-save icon (font awesome "save")
- [x] Save: will update in the existing virtual study if it's mine.
- [x] When you click Save, keep the current behaviour.
- [x] If it already exists, clicking "Save" will update the current study with the current state of the queries. Updating its title won't be possible, for now.
- [x] Use the tooltip text listed below.
- [ ] when there are unsaved changes, should a star (*) and "You have unsaved changes" like in the mockup below.
- [x] disable the save button if no changes have been made.

## Tooltip text
- Save - Saves the current virtual study if it exists, or a new one if not.
- Save as - Saves the current virtual study as a new one.

## Wireframe
Version with the buttons:
![save_as_button](https://user-images.githubusercontent.com/293677/56687935-63b5b300-66a5-11e9-8d7e-843715ceb61f.png)

## Mockup
![image](https://user-images.githubusercontent.com/293677/57714292-34f68100-7642-11e9-8be5-4fb3ef23c292.png)

vferretti:
Each supported action (Save, Edit, etc) requires more precise specifications

aalex:
@jberube Do we already have the endpoints for delete and save as?

jberube:
Not as far as I know. I'll check.

jberube:
Virtual Studies are stored in the Persona API. I don't see a reason for having backend changes, because everything is done in the front-end... which might not be a good idea. We might want to have some more robust architecture, but that may be another subject.

What does "Closing" a virtual study do?

I also need to know the behavior of "edit" and "delete", since we're working shared data, here.

jecos:
Riff has been modified. A pull request is opened in overstack repo https://github.com/overture-stack/riff/pull/21

aalex:
(QA) LGTM, except some bugs that should be verified, and then probably triaged.

PR #1729, opened 2019-05-14
bug/1705 - refactored the SQONProvider to have a more predictable state

jberube:
Introduces a glitch when navigating back to a query that has been loaded, and then modified without being saved.

Will be fixed as part of https://github.com/kids-first/kf-portal-ui/issues/1533

### DIFF
--- a/src/components/CohortBuilder/SQONProvider.js
+++ b/src/components/CohortBuilder/SQONProvider.js
@@ -1,156 +1,201 @@
 import React from 'react';
-import Component from 'react-component-component';
+import PropTypes from 'prop-types';
 import { withRouter } from 'react-router-dom';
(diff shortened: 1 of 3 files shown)


src/components/CohortBuilder/SQONProvider.js
```diff
@@ -1,156 +1,201 @@
...
+  activeIndex: 0,
+  uid: null,
+  virtualStudyId: null,
+};
+
+class SQONProvider extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { ...cloneDeep(initialState) };
```
nyanofthemoon: 😨

nyanofthemoon:
@jberube Fix all the conflicts! :)

PR #1747, opened 2019-05-16
Introducing redux

jberube:
Introducing redux as part of the SQONProvider refactor necessary for #1533 .

Can be merged separately, would be of use to @jaouad-benassila 

This PR only contains the refactor, the Save / Save As will come afterward in a separate PR.


src/store/actionCreators/virtualStudies.js
```diff
@@ -0,0 +1,59 @@
...
+  return dispatch =>
+    getVirtualStudy(api)(virtualStudyId)
+      .then(virtualStudy => {
+        const {
+          uid,
+          content: { sqons, activeIndex },
+        } = virtualStudy;
+
+        dispatch({
+          type: LOAD_VIRTUAL_STUDY,
```
nyanofthemoon: @jberube For good practices, in general I would use the pattern:

- VIRTUAL_STUDY_REQUESTED
- VIRTUAL_STUDY_SUCCESS
- VIRTUAL_STUDY_FAILURE

The main reasons behind this pattern are:

- More leeway in terms of different states in regards to a specific async action
- Easier to understand the flow of the application through the debug tools
- Re-usable across all async actions

src/store/reducers/index.js
```diff
@@ -0,0 +1,7 @@
+import { combineReducers } from 'redux';
+
+import virtualStudies from './virtualStudies';
+
+export default combineReducers({
+  virtualStudies,
```
nyanofthemoon: @jberube I would add a `sqon` reducer as well as this can be re-used across several "pages".

src/index.js
```diff
@@ -15,15 +18,25 @@ initAnalyticsTracking();
...
+const virtualStudies = loadDraftVirtualStudy() || undefined;
+const preloadedState = {
+  virtualStudies,
+};
+const store = initStore(preloadedState);
 
 const render = Component => {
-  ReactDOM.render(<Component />, getAppElement());
+  ReactDOM.render(
+    <Provider store={store}>
```
nyanofthemoon: 👍

src/store/actionCreators/virtualStudies.js
jberube: I agree. I tend to wait until I need at least two of those actions to name them this way, but I don't care about doing that up-front as a general practice for asynchronous actions.

src/index.js
nyanofthemoon: @jberube Provider HOC should only wrap the App Container. There needs to be no state management for static pages.

src/store/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+let store = null;
+export const initStore = preloadedState => {
+  const composeEnhancers = composeWithDevTools({
+    // Specify extension’s options like name,
+    //  actionsBlacklist, actionsCreators, serialize...
+  });
+
+  const enhancer = composeEnhancers(
+    applyMiddleware(thunk),
```
nyanofthemoon: @jberube If we only make simple use of `thunk`, I believe we could remove the dependency and write it as:
```
const thunk = store => {
  const dispatch = store.dispatch
  const getState = store.getState
  return next => action => {
    if (typeof action === 'function') {
      return action(dispatch, getState)
    }
    return next(action)
  }
}
```

src/store/actionCreators/virtualStudies.js
nyanofthemoon: Might as well start this one on the right foot! 👍

src/store/reducers/index.js
jberube: Yes, when time comes.

src/index.js
jberube: That MaintenancePage don't need a store? :rofl: 

Indeed!

src/store/index.js
jberube: I would rather not, for maintenance purpose. Or if we start using TypeScript.

nyanofthemoon: @jberube If we start using TypeScript, let us have a discussion first. I don't want us to start plugging it everywhere, but rather have it in useful places such as validation structs, etc.

PR #1775, opened 2019-05-28
Feat/1533 - Save & Save As

jberube:


package.json
```diff
@@ -56,6 +56,7 @@
     "popback": "^0.0.3",
     "postcss-flexbugs-fixes": "3.2.0",
     "postcss-loader": "2.0.8",
+    "postcss-nested": "^4.1.2",
```
nyanofthemoon: @jberube Why needs?

src/components/CohortBuilder/SaveVirtualStudiesModalContent.js
```diff
@@ -65,7 +69,7 @@ class SaveVirtualStudiesModalContent extends React.Component {
   }
 
   render() {
-    const { name, errorMessage } = this.state;
+    const { name, /*description,*/ errorMessage } = this.state;
```
nyanofthemoon: @jberube Why commented out?

src/components/CohortBuilder/SaveVirtualStudiesModalContent.js
```diff
@@ -95,10 +108,13 @@ const mapStateToProps = state => ({
   loggedInUser: state.user.loggedInUser,
   sqons: state.cohortBuilder.sqons,
   activeSqonIndex: state.cohortBuilder.activeIndex,
+  virtualStudyId: state.cohortBuilder.virtualStudyId,
+  name: state.cohortBuilder.name,
+  description: state.cohortBuilder.description,
```
nyanofthemoon: @jberube Why not disabled?

src/store/reducers/cohortBuilder.js
```diff
@@ -55,6 +73,8 @@ export default (state = initialState, action) => {
       return {
         ...state,
         ...action.payload,
+        dirty: false,
+        areSqonsEmpty: false,
```
nyanofthemoon: @jberube If we reset the virtual study... shouldn't the sqons be empty?

src/store/reducers/cohortBuilder.js
```diff
@@ -55,6 +73,8 @@ export default (state = initialState, action) => {
...
       };
     case VIRTUAL_STUDY_SAVE_FAILURE:
```
nyanofthemoon: @jberube Should it be the same as `default` clause; does this mean if there is a save error the state will lose all values?

aalex:
:+1:

package.json
jberube: To make the & work in:
```
.class-a {
  &.class-b {}
}
``

src/components/CohortBuilder/SaveVirtualStudiesModalContent.js
jberube: Because it's another feature, will be done separately.

I just did the plumbing work while I was at it.

src/store/reducers/cohortBuilder.js
jberube: Yeah, but that is success, not reset.

src/store/reducers/cohortBuilder.js
jberube: No, it won't?

PR #1792, opened 2019-05-30
Update version number and NEWS.md

jberube:


aalex:
TODO: remove fixes done within the features implemented in dev, but never introduced in prod.

PR #1797, opened 2019-05-30
Feature/make most frequent diagnoses chart clickable

jaouad-benassila:


src/components/CohortBuilder/Summary/DiagnosesChart.js
```diff
@@ -75,41 +78,81 @@ const toSingleDiagQueries = ({ topDiagnoses, sqon }) =>
...
-              tickInterval={4}
-              colors={getCohortBarColors(data, theme)}
-              xTickTextLength={28}
-              legends={[
-                { title: 'Probands', color: theme.chartColors.blue },
-                { title: 'Other Participants', color: theme.chartColors.purple },
-              ]}
-            />
-          </BarChartContainer>
+const formatedValue = value => {
```
aalex: This formats the value to pass it to addSqon?

jaouad-benassila: yes because I receive  badly formatted data to use in my sqon

aalex: Yes: so these are fixes needed so that the SQON searches work. Some values cannot be sent back to the backend in the same we received it from it. This doesn't affect what the user sees.

jberube: Shouldn't we just use the `diagnoses__mondo_id_diagnosis` in the query instead?

jaouad-benassila: we use sqon not query , and where do you want to add it ?

PR #1804, opened 2019-05-31
Update NEWS for 2.3.0

aalex:


nyanofthemoon:
Should perhaps rename to `RELEASE_NOTES.md`

aalex:
The GNU standard is to call that file NEWS.

Sometimes, some projects have a file for release notes, but we only leave there the latest release notes. I prefer a leaner approach in which we put all the release notes in anti-chronological order, like I do here.

This file is sometimes called something else. I will let you know if I remember what.

Issue #298, opened 2018-04-19
Implement usersnap or other tool to capture UAT feedback


Fgerthoffert:
This would create issues in a private github repo.

PR #1817, opened 2019-06-06
Add reset zoom to overall survival chart

jaouad-benassila:


src/components/CohortBuilder/Summary/SurvivalChart.js
```diff
@@ -120,7 +131,10 @@ class SurvivalPlot extends React.Component {
           onMouseLeaveDonors:
             this.props.onMouseLeaveDonors && this.props.onMouseLeaveDonors.bind(this),
           onClickDonors: this.props.onClickDonors,
-          onDomainChange: newXDomain => this.updateState({ xDomain: newXDomain }),
+          onDomainChange: newXDomain => {
+            this.props.disableResetButton() && this.props.disableResetButton.bind(this);
```
jberube: Can you explain what this line does?

jaouad-benassila: to disable  reset zoom button , I think is clear

jberube: Yeah, what it does its clear.

What preocupies me is the second part of the line: 
` && this.props.disableResetButton.bind(this);`

From what I understand, `this.props.disableResetButton()` will be `handleDisableResetButton` at line #298 which never returns any value:
```
  handleDisableResetButton = () => {
    this.setState({ zoomDisabled: false });
  };
```

So why the ` this.props.disableResetButton() && this.props.disableResetButton.bind(this);` and not just `this.props.disableResetButton()`?

aalex:
I had to use my admin priviledges to merge this PR. @nyanofthemoon recommended to merge it. Please create a refactoring backlog item if needed. (to address the changes needed according to the conversation around this PR)

PR #1837, opened 2019-06-13
Feat/916 participant summary

Velythyl:
Does the summary page.

NOTES: 
We cut out the Number Of Files for a biospecimen as it requires multiple queries and then a join between the JSONs in JS for that single table cell.

Of particular note is the "getter" function in ParticipantSummar.js. Sometimes, nodes are simply missing, and directly accessing the fields crashes the pages with an Exception. I find that function very ugly; was this the "best" way to handle this corner-case?


src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -0,0 +1,103 @@
...
+          //uses a fake Link: we want the same style as a SecondaryNavMenu, and this is the key to it
+          return (
+            <li
+              style={
+                {
+                  borderRight: (isActive || this.state.hovered) ? "5px solid #e83a9c" : "5px solid transparent"
+                }
+              }
+
+              onClick={ () => holder.setState({activeTab: tabId})}
```
jberube: You should use `this`. To have `this` being bound to the correct value, use `bind`.

i.e. `this.makeTabButtons = this.makeTabButtons.bind(this)` in the constructor.

If you end up doing that more than once in the same class, use the `autobind` instead:
```
import autobind from 'auto-bind-es5';
...
constructor(props) {
  super(props);
   ...
  autobind(this);
}
```

It basically does the above on every method of an object. It's used in a few places in this project.

src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -0,0 +1,103 @@
...
+   * @param holder This, as it is needed inside a nested class
+   * @returns {*|Array|*[]}
+   */
+  makeTabButtons(holder) {
+    return this.tabIDs.map(tabId => {
+
+      const isActive = tabId === this.state.activeTab;
+
+      //needs to be a class to use setState to do the hovering. Would do in CSS, but with emotion it's too complicated and hacky
+      class Button extends React.Component {
```
jberube: Move the class outside of the `makeTabButtons` function, please, and pass `isActive` and `tabId` as props.

Classes should always be declared at the root of the module to avoid nasty nasty hoisting bugs.

src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -0,0 +1,103 @@
...
+   *
+   * @param holder This, as it is needed inside a nested class
+   * @returns {*|Array|*[]}
+   */
+  makeTabButtons(holder) {
+    return this.tabIDs.map(tabId => {
+
+      const isActive = tabId === this.state.activeTab;
+
+      //needs to be a class to use setState to do the hovering. Would do in CSS, but with emotion it's too complicated and hacky
```
jberube: You can also use css at the same time.

see `src/components/CohortBuilder/VirtualStudiesMenu/index.js:36` and `src/components/CohortBuilder/VirtualStudiesMenu/index.postcss`. You can do the same with plain `.css` file if you prefer.

Just remember to use `className` and not `class` when setting a class in react.

src/uikit/SecondaryNav/SecondaryNavMenu.js
```diff
@@ -6,9 +6,9 @@ import { isEmpty } from 'lodash';
 
 import SecondaryNavTab from './SecondaryNavTab';
 
-const MenuWrapper = styled('ul')`
-  ${({ theme }) => theme.secondaryNav}
-`;
+  const MenuWrapper = styled('ul')`
```
jberube: That should not be indented.

src/uikit/SummaryTable/VariableSummaryTable/Tables.js
```diff
@@ -34,9 +34,32 @@ const SummaryContent = styled('div')`
...
+ * @param amount The divider (the number of chunks we want)
+ * @returns {number} The biggest chunk.
+ */
+function splitter(number, amount) {
+  const approx = number / amount;
+
+  if(Number.isInteger(approx)) return approx;
+
+  const integer = Math.floor(approx);
+  const decimal = "0."+(approx+'').split('.')[1]; //weird, but best way to get precise decimal places https://stackoverflow.com/questions/4512306/get-decimal-portion-of-a-number-with-javascript
```
jberube: Sad but true. Damn those JS unreliable floats. `const decimal = approx % 1;` would be so elegant!

src/uikit/SummaryTable/VariableSummaryTable/index.js
```diff
@@ -0,0 +1,55 @@
...
+      align-self: flex-start;
+    }
+
+    > div:last-child {
+      margin-right: 0;
+    }
+  }
+`;
+
+const VariableSummaryTable = ({ rows, nbOfTables }) => (
```
jberube: 👍 Very elegant!

src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -0,0 +1,103 @@
...
+  ${({ theme }) => theme.secondaryNav}
+`;
+
+class Holder extends React.Component {
+
+  constructor(props) {
+    super(props);
+
+    //wrap the item in a one-element array if the item is not itself an array
+    if(!(props.children instanceof Array)) {
```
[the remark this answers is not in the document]
jberube: Also, since the values of `this.tabIDs` and `this.children` are calculated from the `props`, they should be recalculated each time the props change.

Easiest way is to calculate them in the `render`, but if you need them outside of the render, you can use the lifecycle event https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops , and end-up using the state instead of properties on `this`.

src/components/EntityPage/Participant/ParticipantSummary.js
```diff
@@ -1,32 +1,182 @@
...
+ *
+ * We're taking the opportunity to sannitize the return value:
+ *  null = --
+ *  boolean = true/false
+ *
+ * @param obj
+ * @param accessor
+ * @returns {string|*|null|undefined}
+ */
+function getter(obj, accessor) {
```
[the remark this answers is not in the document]
jberube: Only when accessing a property of a property that is an object (including arrays and functions), to avoid `TypeError: can't access property "x" of undefined`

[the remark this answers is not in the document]
jberube: Also, It's better to use gt this way:
```
import { get } from 'lodash';

get(...);
```
rather than:
```
import * as _ from 'lodash';
or 
import _ from 'lodash';

_.get(...);
```

This is due to the treeshaking done by Webpack. Webpack won't have to import the whole lodash with the first method: it may have to with the second. It helps to make a smaller package in the end.

src/components/EntityPage/Participant/Utils/Holder.js
jberube: I will need to run it locally to see what can be done. I'll do that after you push.

src/components/EntityPage/Participant/Utils/Holder.js
jberube: I double checked and I am wrong. Classes are not hoisted.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/class#Description

Yet, I'd rather have classes being declared at the top level of the module. It can't be used from outside the module if you don't `export` it.

src/components/EntityPage/Participant/Utils/Holder.js
jberube: You should not have the need to. Affecting the state of a parent directly is a big no-no.

See: https://reactpatterns.com/#state-hoisting

jberube: Also forgot to mentioned: you should add a `key` property your Button, which should:

- be unique to that button within it's siblings
- never change

That allows react to track siblings that change order or added / removed elements, to optimize it's performance and allow animation framework to track elements.

In this case it's not useful, but react will add a warning in the console and we try to eliminate them.

https://reactjs.org/docs/lists-and-keys.html

src/components/EntityPage/Participant/ParticipantSummary.js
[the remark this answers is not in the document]
jberube: Use `get` and provide a default value (third parameter) which is of the same type as you would get otherwise, so an empty string, empty array, etc...

In cases of Numbers, your return value can depend on what makes more sense to render. Maybe not having a number means `0`, maybe a "no value" indicator e.g. `--` make more sense.

src/components/EntityPage/Participant/Utils/Holder.js
Velythyl: Okay! It's done.

src/components/EntityPage/Participant/Utils/Holder.js
Velythyl: Okay, changed it to a callback! And I added the keys

src/components/EntityPage/Participant/Utils/Holder.js
Velythyl: Okay! I'll Slack you when i push

src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -1,99 +1,108 @@
...
+    //wrap the item in a one-element array if the item is not itself an array
+    if(!Array.isArray(this.props.children)) {
+      tabIDs = [this.props.children.props.label];
+      children = [this.props.children];
+    } else {
+      tabIDs = this.props.children.map( child => child.props.label );
+      children = this.props.children;
+    }
+
+    window.console.log(children);
```
jberube: Please remove `console.log`.

src/components/EntityPage/Participant/Utils/Holder.js
```diff
@@ -1,99 +1,108 @@
...
+}
+
+class Holder extends React.Component {
+
+  constructor(props) {
+    super(props);
 
-      return <Button />
-    })
+    this.state = {activeTab: (Array.isArray(props.children) ? props.children[0].props.label : props.children.props.label)};
```
jberube: Leave it like that now that it works, but I think it would have been better to keep an index here, to avoid having to read the props of children.

PR #1841, opened 2019-06-14
Proposal of a coding style for the UI code

aalex:
Please try it out, and suggest changes to it, as needed.


jberube:
We already have 3 tools that touch linting and formatting:
- eslint
- prettier
- tslint (through babel)

We already have a prettier configuration in package.json, the one you propose in incompatible with it.

Where does the need come from?

aalex:
Good to know! I hadn't seen the other one. I made this one. I'll close this PR. The source code documentation should probably point to those resources, somewhere.

PR #1844, opened 2019-06-14
Add tab persistence to cohort builder

jaouad-benassila:


src/store/reducers/cohortBuilder.js
```diff
@@ -117,6 +119,12 @@ export default (state = initialState, action) => {
     case LOGOUT:
       return cloneDeep(initialState);
 
+    case SET_ACTIVE_VIEW:
+      return dirty({
```
jberube: Don't use `dirty` here, it flags the virtual study has having changed, and this is not the case.

PR #1848, opened 2019-06-18
clinical page done. 

Velythyl:
Could not do Phenotypes as they are not in grapql (broken?). Possible improvements: add phenotypes. Add diagnoses and outcomes to family members (also not in graphql (broken?)).


src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -1,15 +1,46 @@
...
+  //const phenotypes = getNodes(participant, "phenotype", []);
 
-const ParticipantClinical = () => {
   return (
     <React.Fragment>
-      <EntityContentSection title="Diagnoses">Diagnoses Chart</EntityContentSection>
+      <EntityContentSection title="Diagnoses">
+        {diagnoses.length === 0 ?
+          <div>No diagnoses.</div> :
+          <ControlledDataTable loading={false} columns={diagHeads} data={diagnoses} dataTotalCount={-1} onFetchData={() => null} />
```
jberube: Sorry about that, `ControlledDataTable` should have allowed to use it in a "static" way more easily.

You could pass it a `manual={false}` property, and pass that property to the `ReactTable` component. Just be sure to default the value of `manual` inside `ControlledDataTable` to `false`.

src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -1,15 +1,46 @@
...
+        }
+      </EntityContentSection>
       <EntityContentDivider />
-      <EntityContentSection title="Phenotypes">Phenotypes Chart</EntityContentSection>
+      <EntityContentSection title={"Family"}>
+        <div style={{color: "#404c9a", fontWeight: "bold"}}>
+          <img src={require("../../../assets/icon-families-grey.svg")} style={{height: "1em", marginRight: "1em"}} alt={"family icon"}/>
```
jberube: You can move that reuiqre to the top of the file and use an `import` instead. It will give a you the same result more easily, as you can use aliases in the path. It's also more refactor friendly.

```
import familyIconPath from 'assets/icon-families-grey.svg';
```

src/components/EntityPage/Participant/Utils/sanitize.js
```diff
@@ -0,0 +1,22 @@
+/**
```
jberube: ❤️ 
Nice function!

Just be extra careful not to mutate the parameter: you might have a nasty surprise if you pass the value to another function that expect the value _not_ to be modified.

Those functions should return new values. Either ensure the immutability of the function, or clone it's input before mutating it. Since that one is recursive, cloning you have a bigger impact, so I would suggest just to keep immutability.

If I'm not clear (sorry, in a hurry), I'll show you tomorrow what I mean.

src/components/EntityPage/Participant/index.js
```diff
@@ -53,6 +52,11 @@ class ParticipantEntity extends React.Component {
   render() {
     const { participantId, location, participant, isLoading, error } = this.props;
 
+    //only way to update the page when we click a link to a participant in clinical...
+    if(this.props.participant !== null && this.props.participant.kf_id !== this.props.participantId) {
```
jberube: Should you check isLoading too to prevent sending the query many times?

src/components/EntityPage/Participant/Utils/sanitize.js
jberube: Here's an example of a way to achieve this with `Object.prototype.keys` and `Map.prototype.reduce`:

```js
function sanitize(obj) {
  if (Array.isArray(obj)) return obj.map(sanitize);

  return Object.keys(obj).reduce((newObj, key) => {
    const val = obj[key];
    newObj[key] = val === null ? '--' : typeof val === 'boolean' ? `${val}` : obj[key];
    return newObj;
  }, {});
}
```

I haven't tested, though.

src/components/EntityPage/Participant/ParticipantClinical.js
Velythyl: done!

src/components/EntityPage/Participant/Utils/sanitize.js
Velythyl: Alright! I tested it and I works ^^

I didn't know that JS had reduce, I'll keep that in mind for the future :) Thanks!

src/components/EntityPage/Participant/index.js
Velythyl: From what I've tested the check to see if this.props.participant is null handles that, but I can add a check to isLoading to be sure!

src/components/EntityPage/Participant/ParticipantClinical.js
Velythyl: I'm not sure what you mean :/ I looked at the source of ControlleDataTable and the "fake" props I'm passing it (loading, dataTotalCount and onFetchData) are all required. I feel like if we're adding a `manual` prop we'd have to remove the propTypes that are there right now.

jberube: You're right, it would need a little more refactor.
Let's forget about this.

src/services/profiles.js
```diff
@@ -61,6 +61,9 @@ export const getProfile = api => async () => {
     },
   });
 
+  console.log(url)
```
jberube: oupsy!

Issue #1722, opened 2019-05-10
Saved queries: show the description of each virtual study


aalex:
As a user, I want to see a widget that lists my virtual studies.

(this user story is part of an epic - see below)

## Acceptance criteria
- Show the description for each virtual study.
- This text must reflect the actual description for each virtual study and be kept up-to-date.
- Limit to 60 characters. Use an ellipsis if its more than that.
- Imitate a "pre" tag - in which multiple white spaces are shown as such. (imitate the behaviour of the description in the cohort builder page, for that)
- Add a tooltip that show the full description of the virtual study when you mouse over it.
- Use the same grey text font and style as the other tab. (not in bold)

## Mockups
![image](https://user-images.githubusercontent.com/293677/57876868-69a23e00-77e4-11e9-8f64-f689d0ac6e83.png)

## Technical discussion
- It should be implemented better than in the file repository. In the file repository, it's calculated once and then stored. Here, we want to dynamically query it each time the page is loaded.

aalex:
(QA) The PR attached to this item has been closed, but the branch has not been merged. The feature is therefore not implemented in the dev/QA environment. Someone should then continue the work in that branch (or in a new branch) and then open a new PR.

jberube:
I am on it at the moment, thanks.

aalex:
The PR https://github.com/kids-first/kf-portal-ui/pull/1861 had been linked to this item, but then got closed without being merged.

aalex:
(QA) LGTM

![image](https://user-images.githubusercontent.com/293677/60686335-165c7b80-9e76-11e9-88af-9e21fa497937.png)

aalex:
(QA) Changes requested:
- Display the description of the virtual study on more than one line, if needed. 
- Increase the maximum length of the description to 140 characters.
- Make sure there is no text under the "Delete" button. (the trashcan)

aalex:
(QA) LGTM

![image](https://user-images.githubusercontent.com/293677/60919658-53ed4a00-a264-11e9-8752-8cf8d620561e.png)

PR #1861, opened 2019-06-28
Show description of each virtual study

jaouad-benassila:


src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -191,10 +225,18 @@ export const MySavedQueries = compose(
...
                         <Scroll>
-                          {virtualStudies.map(s => (
-                            <Study>
+                          {virtualStudies.map((s, index) => (
+                            <Study key={s.id}>
                               <Row justifyContent="space-between" width="100%">
-                                <StudyLink to={`/explore?id=${s.id}`}>{s.name}</StudyLink>
+                                <div className={`${studyStyle}`}>
+                                  <StudyLink to={`/explore?id=${s.id}`}>{s.name}</StudyLink>
+                                  {virtualStudyLoad(s.id, index, props)}
```
jberube: That cause a query for every virtual studies of the user at each render.

Please move to the appropriate lifecyle event.

src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -78,6 +99,19 @@ const mapStateToProps = state => {
   };
 };
 
+let descriptions = new Array(20);
```
jberube: Why did you use a module variable here? It won't be set when the query returns, so there won't be a render unless something else triggers it.

Please move to the state of the component, or use redux to store the virtual studies info. We currently only store the information of the loaded study in the cohort-builder, but we could store them all now that there is a need for it. See `src\store\reducers\ui\participantEntityPage.js` for an example. Reducers in the `ui` folder are for specific parts of the UI, if we end-up re-using the information, we will move it to a general-purpose reducer later.

jberube:
I've found an endpoint of Riff that allows to get the whole content of a user, that would be more effective than querying for each saved study: 
https://github.com/kids-first/riff/blob/master/src/main/java/bio/overture/riff/controller/RiffController.java#L50

aalex:
This PR had been connected to https://github.com/kids-first/kf-portal-ui/issues/1722 - but the PR has been closed without being merged.

PR #1894, opened 2019-07-05
Change the link of studies and participants clickable diagrams in the…

nyanofthemoon:


src/components/UserDashboard/charts.js
```diff
@@ -175,23 +164,41 @@ export const UserInterestsChart = withTheme(({ data, theme }) => {
...
+)(({ data, theme, setSqons, virtualStudy, history }) => {
 
   const onClick = barData => {
     trackBarClick(diagnosesChartCategory, barData);
-    history.push(barData.data.url);
+    addSqon('diagnoses.diagnosis', barData.data.name);
+    history.push('/explore');
+  };
+
+  const addSqon = (field, value) => {
```
Velythyl: Small thing: would extracting the addSqon function and making it top-level in the file a good idea? The addSqon function is also used at line 88, so we could avoid duplicated code this way. I don't know if this is the right move though as addSqon has side effects :)

nyanofthemoon: Would rather not since the components are now connected to the state.

Velythyl: (passing the virtualstudy as a parameter would be required, though)

Velythyl: Alright! LGTM then :)

PR #1914, opened 2019-07-09
Fixed and ready to merge

Velythyl:
changed wrapping css to word-break. Added div and the same style to tooltip. Number of chars now calculated in JS instead of CSS


src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -76,10 +76,7 @@ const Scroll = styled('div')`
 const studyDescriptionStyle = css({
   fontSize: '12px',
   fontFamily: 'Open Sans, sans-serif',
-  maxWidth: '400px',
-  textOverflow: 'ellipsis',
-  whiteSpace: 'nowrap',
-  overflow: 'hidden',
+  wordBreak: 'break-word',
```
nyanofthemoon: 🦄

PR #1931, opened 2019-07-10
fixed 1681 by changing the formatter for the axis to ignore decimals

Velythyl:


src/chartkit/components/HorizontalBar.js
```diff
@@ -139,7 +139,7 @@ class HorizontalBar extends Component {
       indexBy = 'id',
       height,
       tooltipFormatter,
-      axisBottomFormat = v => v.toLocaleString(),
+      axisBottomFormat = v => Number.isInteger(Number(v)) ? v.toLocaleString() : "",
```
nyanofthemoon: @Velythyl Where should the data for 0.5 or 1.5 go?

Velythyl: it won't be displayed.

From what I can see, the only uses of those Bars are to display numbers of stuff that only make sense in integers

Like participants, or number of files, or number of diagnoses

PR #1985, opened 2019-07-29
added some missing information in NEWS.md

jberube:
Also prettified the file.


jberube:
Will be updated on another PR instead.

PR #1993, opened 2019-08-05
fix/1961 - access of DCF files not taken into account properly

jberube:
Acl of DCF files were not taken into account.

⚠️  Needs to be tested before this fix is accepted.

Also, it had to be based on `fix/1991-file-entity-phenotypes` because the file entity page wasn't working . Will be rebased on dev once `fix/1991-file-entity-phenotypes` is merged.


nyanofthemoon:
@jberube Cannot test.
image

nyanofthemoon:
@jberube works on my local connected to prod env.

PR #2025, opened 2019-08-19
fix/1961 - fixed 'download' buttons for open access files in file rep…

jberube:
Fixed 'download' buttons for open access files in file repository table and file entity page.

### DIFF
--- a/src/services/fence.js
+++ b/src/services/fence.js
@@ -137,7 +137,8 @@ export const downloadFileFromFence = async ({ fileUuid, api, fence }) => {
   try {
     accessToken = await getAccessToken(api, fence);
   } catch (err) {
-    console.warn(`[Fence] no access token fo file ${fileUuid}, assuming open access`);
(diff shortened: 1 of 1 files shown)


src/services/fence.js
```diff
@@ -133,12 +133,21 @@ export const deleteFenceTokens = async (api, fence) => {
  * Download File From a Fence Data Repository
  */
 export const downloadFileFromFence = async ({ fileUuid, api, fence }) => {
-  let accessToken = await getAccessToken(api, fence);
+  let accessToken = null;
+  try {
+    accessToken = await getAccessToken(api, fence);
+  } catch (err) {
+    console.warn(`[Fence] no access token fo file ${fileUuid}, assuming open access`);
```
jecos: Typo 'fo' -> 'to'
Should we mentionned the file uuid in the error message, because access token is not related to the file?

PR #2051, opened 2019-08-29
2048 : changed gen3 logo for kf framwork services

evans-g-crsj:


evans-g-crsj:
TODOs :

-  remove gen3Logo after making sure we won't reuse it;

- see if the new logo breaks in older browser we must support.

Issue #1920, opened 2019-07-09
Member page: various fixes to do


aalex:
As a user, want to see a member's page, and be able to edit mine.

There are a few problems with the settings/member page that are obvious, and probably easy to fix.

## Bugs description
- Wrong email: The mailto link is always the one that was on the original mockup. Instead, it should point to the member's email address.
- Remove capitalization for the institution field.
- Remove capitalization in the address.
- The special characters in the phone number should not be removed. If I enter "+1-514-522-2222, ext 123", it shows up as "15145222222123". It should not.
- If the role of the member is patient/family or community member: ("patient" or "community")
    - hide their jobTitle
    - hide their institution
    - hide their department

## Screenshots
Wrong email:
![image.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/86bedc3b-1a70-47fb-a1b4-fe8f3c7ca1b5)

Remove capitalization for the institution field
![image.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/bff28ae6-f9c5-4105-b573-4b1f2f12e203)

Remove capitalization in the address
![image](https://user-images.githubusercontent.com/293677/60916600-d245ee00-a25c-11e9-928f-b5fb3aa5d3e7.png)

The special characters in the phone number should not be removed.
![image.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/85298fe2-08fd-4550-83d6-b27174e3fac9)

 If the role of the member is patient/family or community member hide their institution, etc.
![image.png](https://images.zenhubusercontent.com/5c55147e9ec3935c16660666/ccfa5e9c-42b6-43ff-969f-4db154880ae1)

PR #2056, opened 2019-09-04
#1920 bug fixes

evans-g-crsj:
I removed the phone number formatting completely. I think we should enhance the UI for better control over it (like adding a box for the extension) and/or use an appropriate library to deal with all the different phone number patterns.

I noticed that there is not input validation (for instance, the email could be something like : baljslfdjlk3 and still be accepted).


evans-g-crsj:
apparently, these fixes were already made in another branch. So, I ll make sure it s the case in close this pr for now.

PR #2073, opened 2019-09-18
Feature/2070 modify dx table

evans-g-crsj:
Side note not related to the feature: while working on this page I noticed that sometimes it displays one table and a few seconds later one or two other table appears (qa was slow at that time). There is no loader indicating that a data fetching is occurring.

### DIFF
--- a/src/components/EntityPage/Participant/ParticipantClinical.js
+++ b/src/components/EntityPage/Participant/ParticipantClinical.js
@@ -19,20 +19,22 @@ import { resetVirtualStudy } from '../../../store/actionCreators/virtualStudies'
 import { store } from '../../../store';
 import prettifyAge from './Utils/prettifyAge';
 import { flatMap } from 'lodash/collection';
-import { HPOLink, SNOMEDLink , MONDOLink, NCITLink} from '../../Utils/DiagnosisAndPhenotypeLinks';
-
+import { HPOLink, SNOMEDLink, MONDOLink, NCITLink } from '../../Utils/DiagnosisAndPhenotypeLinks';
+import ClinicalIcon from 'icons/ClinicalIcon';
+import BiospecimenIcon from 'icons/BiospecimenIcon';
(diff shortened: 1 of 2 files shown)


src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -43,8 +46,8 @@ class ParticipantClinical extends React.Component {
       });
     }
 
-    let diagnoses = get(this.props.participant, 'diagnoses.hits.edges', []).map(ele =>
-      Object.assign({}, get(ele, 'node', {})) //copy obj
+    let diagnoses = get(this.props.participant, 'diagnoses.hits.edges', []).map(
+      ele => Object.assign({}, get(ele, 'node', {})), //copy obj
```
jecos: Remove comma

jberube: Auto-format by Prettify.

It makes the code uniform, but it sometimes makes a :hankey:

evans-g-crsj: good point

src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -254,5 +323,8 @@ class ParticipantClinical extends React.Component {
     );
   }
 }
-//: {get(participant, "family.family_composition.hits.edges[0].node.composition", "trio")}
-export default withRouter(ParticipantClinical);
+
+const ParticipantClinicalWithRouter = withRouter(ParticipantClinical);
+const ParticipantClinicalWithRouterAndTheme = withTheme(ParticipantClinicalWithRouter);
+
+export default ParticipantClinicalWithRouterAndTheme;
```
jberube: You can simply do (if you prefer):
```js
export default withTheme(withRouter(ParticipantClinical));
```

PR #2076, opened 2019-09-18
fixed typo introduced by me (0 participantS), added cursor to biospec…

evans-g-crsj:
…imen chip, merged 'type' cell into 'category' cell, added tooltip to icons and formatted biospecimens ids.

### DIFF
--- a/src/components/EntityPage/Participant/ParticipantClinical.js
+++ b/src/components/EntityPage/Participant/ParticipantClinical.js
@@ -24,6 +24,7 @@ import ClinicalIcon from 'icons/ClinicalIcon';
 import BiospecimenIcon from 'icons/BiospecimenIcon';
 import { withTheme } from 'emotion-theming';
 import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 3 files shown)


src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -154,7 +155,7 @@ class ParticipantClinical extends React.Component {
   }
 
   showParticipantNb = (wrapper, field, value) => {
-    if (wrapper.value === '0 participants') return <div>0 participant</div>;
+    if (wrapper.value === '0 participants') return <div>0 participants</div>;
```
jecos: do you really need a specific rule for 0 participant?

jecos: Actually, yes we need thius. It replace the link by a div

evans-g-crsj: as the code stands right now, I would say yes.

PR #2092, opened 2019-09-30
2021/family table remove not shared data

evans-g-crsj:


src/components/EntityPage/Participant/Utils/FamilyTable.js
```diff
@@ -3,232 +3,249 @@ import { get, flatMap } from 'lodash';
...
+const buildDefaultForMembers = (ids = [], defaultValue = '--') => {
+  return ids.reduce((acc, id) => {
+    return {
+      ...acc,
+      [id]: defaultValue,
+    };
+  }, {});
+};
+
+const baselineWithIds = (leftfield, accessor = '', subheader = false, subaccessor = '') => ids => {
```
jberube: I think it would make more sense to change the signature to:
```js
const baselineWithIds = ids => (leftfield, accessor = '', subheader = false, subaccessor = '') => {
```
since you always call it from the same function with the same value for `ids`. Just store `baselineWithIds(allIds)` in a variable at the beginning of `buildRows`, you'll remove a lot of function calls from the render path.

evans-g-crsj: Good point.

PR #2094, opened 2019-10-03
1) Make sure that the family table displays only when the data are sh…

evans-g-crsj:
…ared among family members. 2) Show a msg "No Clinical Data Reported" when there is no table to display. 3) cleanup + show loader when fetching data.


src/components/EntityPage/Participant/ParticipantClinical.js
```diff
@@ -25,20 +25,123 @@ import BiospecimenIcon from 'icons/BiospecimenIcon';
 import { withTheme } from 'emotion-theming';
 import isEmpty from 'lodash/isEmpty';
 import Tooltip from 'uikit/Tooltip';
-//https://kf-qa.netlify.com/participant/PT_C954K04Y#summary tons of phenotypes
-//https://kf-qa.netlify.com/participant/PT_CB55W43A#clinical family has mother and child being affected
```
jecos: This is multi-line comment on a multi-line comment 🤣

PR #2097, opened 2019-10-08
remove protection on fences + make sure that file data is passed in e…

evans-g-crsj:
…very scenario where only one file is choosen.

### DIFF
--- a/src/components/EntityPage/File/index.js
+++ b/src/components/EntityPage/File/index.js
@@ -315,6 +315,11 @@ const FileEntity = compose(withTheme)(
                 fileId={fileId}
                 disabled={!hasFilePermission}
                 hasFilePermission={hasFilePermission}
+                file={{
+                  acl: data.acl,
+                  latest_did: data.latest_did,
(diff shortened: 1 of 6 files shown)


src/components/EntityPage/File/index.js
```diff
@@ -315,6 +315,11 @@ const FileEntity = compose(withTheme)(
                 fileId={fileId}
```
jecos: fileId could be part of file

src/components/FileRepo/CustomColumns/ActionsColumn.js
```diff
@@ -140,6 +141,7 @@ const ActionsColumn = ({ value, api, theme, fenceAcls }) => (
           const acl = file.acl || [];
           const repository = file.repository;
           const hasAccess = acl.includes('*') || intersection(fenceAcls, acl).length > 0;
+          console.log('file ', file, 'data ', data); //==========================================================================TODO
```
jecos: remove

PR #2117, opened 2019-10-16
fix some warnings

evans-g-crsj:


src/utils.js
```diff
@@ -0,0 +1,13 @@
...
+  which makes it almost impossible at the moment to fix them correctly.
+  So, for the sake of depolluting the warnings, I'll use the following method. 
+*/
+
+export const removeUnwantedProps = (incomingProps = {}, blackListedPropKeys = []) => {
+  const filteredEntries = Object.entries(incomingProps).filter(
+    ([key]) => !blackListedPropKeys.includes(key),
+  );
+  return Object.fromEntries(filteredEntries);
+};
```
jberube: Nice implementation of https://lodash.com/docs/4.17.15#omit :stuck_out_tongue: 

We might want to go with `lodash/omit`.

evans-g-crsj: True. I'll do that!

PR #2118, opened 2019-10-16
Feature/1881/list member page

adipaul1981:


src/components/MemberSearchPage/MemberTable.js
```diff
@@ -0,0 +1,137 @@
...
+    return (
+      <div key={index} style={{ paddingRight: 5, wordWrap: 'break-word', maxWidth: '100%' }}>
+        {value}
+      </div>
+    );
+  }
+  for (const h of highLightValues) {
+    if (value === h.replace(regex, '')) {
+      const arr = h.split(regex);
+      if (arr.length >= 3) {
```
evans-g-crsj: magic number

src/index.css
```diff
@@ -31,6 +31,11 @@ a.primary {
   color: #90278e;
 }
 
+#grid-container{
```
evans-g-crsj: question : should we put all of our css in one file or use feature-specific file (e.g. one css page for members, one css page for file repos ,...)? If only one file is used (like it is now), I think we should rename #grid-container with something like #members-list-grid-container or something like that..

adipaul1981: yes. I've put them in an component package specific css file.

src/components/MemberSearchPage/MemberTable.js
```diff
@@ -0,0 +1,143 @@
...
+  if (!highLightValues) {
+    return (
+      <div key={index} className={'format-label'}>
+        {value}
+      </div>
+    );
+  }
+
+  const res = [];
+  highLightValues.forEach(h => {
```
evans-g-crsj: Not sure index is correct, here. It should be the index of the forEach loop. I would go with "highLightValues.filter().map()"

adipaul1981: You are correct about the index. I'm changing it. Your other comment about "highLightValues.filter().map" actually point out a mistake on my part. I'm actually discarding with the if condition the interests that are not a "Hit". Your comment made me realize that there is a mistake in my code.  The "forEach" is justified in this case i think. I just need to add an "else" condition to my "if" statement and keep the one not a "hit" un-formatted.

adipaul1981: humm... actual on second thought the original index was correct. It is not the index of the forEach i should use. But the original index of the passes to the component.

Issue #2031, opened 2019-08-22
Download Clinical Data/Biospecimen Data


jberube:
![image](https://user-images.githubusercontent.com/12585875/64203192-effc9180-ce60-11e9-888c-530ae409374f.png)

# Clinical Data

> Note: currently, in the downloaded data, we have no data for the mother and the father (display: --). For the proband column, we have either true or nothing (noted -): we must be able to put true or false. Finally, for the Study Name, the study short name field will be displayed.

### Point to improve

The file format obtained when downloading participants' clinical data must be optimized (in CB and FR). For that, we choose to download an excel file with several tabs (or a google sheet?)

### Model:
Here is the model of the Excel file to implement:
https://docs.google.com/spreadsheets/d/16lRjd_h8FnNKLAoVWw5Li6o4pd_eyysTJeAS6MzHXBo/edit?ts=5d6430fe#gid=0

## Participants & Family Members
*We need to discuss the format* but it will be very similar to the participants with, in addition, the following columns:
- the identifier of the family
- The relationship with the proband (?)

# Biospecimen Data
*Its format will probably also be modified*

PR #2122, opened 2019-10-21
feat/2031 - Download Clinical Data/Biospecimen Data

jberube:
The client side of #2031 .

Server-side: https://github.com/kids-first/kf-api-portal-reports/pull/2

Some early feedback would be appreciated!

### DIFF
--- a/src/common/downloader.js
+++ b/src/common/downloader.js
@@ -1,0 +1,87 @@
+import axios from 'axios';
+import { saveAs } from 'file-saver';
+import uuid from 'uuid';
+
+const getDefaultContentType = responseType => {
+  const fallbackContentType = 'text/plain;charset=utf-8';
+
+  switch (responseType) {
+    case 'json':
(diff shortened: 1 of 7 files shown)


src/common/downloader.js
```diff
@@ -0,0 +1,91 @@
...
+          const filenameFromContentDisposition = disposition
+            .split(';')
+            .map(part => part.trim())
+            // Matching the modern format`filename="MyFile.txt"`,
+            //  but not the derelict syntax `filename*=utf-8''MyFile.txt`
+            .find(part => /filename=".*"/i.test(part));
+          filename = /filename="(.*)"/i.exec(filenameFromContentDisposition)[1];
+        } catch (err) {
+          filename = uuid();
+          console.warn('failed to parse filename, will fallback to an UUID', disposition);
```
evans-g-crsj: Should we notify the user?

PR #2128, opened 2019-10-24
Feature/1881 list members qa

adipaul1981:
1. Added the background
2. Fix CSS for title (temporary fix)
3. Other general CCS fixed (minor)
4. Interests display: remove the more/close toggle and replaced it with a one time only click + icon


src/components/MemberSearchPage/MemberIntersts.js
```diff
@@ -30,6 +30,10 @@ class MemberInterests extends Component {
       filter: !prevState.filter,
     }));
   }
+  @bind
```
jecos: should be removed?

adipaul1981: yes i seen it a couple of minutes after submitting the PR...

PR #2135, opened 2019-10-25
manage more cavatica errors

evans-g-crsj:


src/components/cavatica/CavaticaAddProject.js
```diff
@@ -56,84 +57,140 @@ const AddIcon = styled(PlusIcon)`
...
+        addingProject: false,
+      });
+    } catch (error) {
+      this.setState({ error });
+    } finally {
+      this.setState({ projectName: '', isSaveButtonDisabled: false });
+    }
+  }
+
+  @bind
```
jberube: We'll have to stop using decorators as it will not be supported by CRA until the API is stable.

Generally speaking, we should refrain from using ES features that do not have a stable spec (at least stage 4).

jberube: See https://github.com/facebook/create-react-app/issues/411

PR #2141, opened 2019-10-29
add featureToggle + bug fix

evans-g-crsj:


src/components/UserProfile/Account.js
```diff
@@ -65,45 +66,47 @@ export default compose(
...
-              onClick={checked => {
-                profile.isPublic = checked;
-                submit(profile);
-              }}
-            />
-          }
-        />
-      </IntegrationTable>
-    </SettingsSection>
+    {isFeatureEnabled('searchMembers') && ( //TODO : remove me one day :)
```
jberube: Just throwing the idea here, not requesting a change.
What do you think of the idea of a HOC to deal with feature toggles?
i.e. 
```jsx
<FeatureToggle enabled="searchMembers">
  <SettingsSection>
    ...
```

src/common/featuresToggles.js
```diff
@@ -1,22 +1,14 @@
...
-  const flag = getApplicationEnvVar(featureName);
-
-  // missing feature flags defaults to their default
-  if (typeof flag === 'undefined') return defaultValue;
-
+const isEnabled = flag => {
   // any casing for 'false', false, 0 and '0' are all disabled
   // eslint-disable-next-line eqeqeq
-  if (flag === false || flag.toString().toLowerCase() === 'false' || flag == 0) {
+  if (!flag || flag.toString().toLowerCase() === 'false' || flag == 0) {
```
jberube: With that change, it would be a little more readable to change `flag == 0` for `flag === '0'` and get rid of `eslint-disable-next-line eqeqeq`.

evans-g-crsj: yes, I'll change that

PR #2142, opened 2019-10-31
1042 fix login bump when already connected

evans-g-crsj:


src/App.js
```diff
@@ -40,6 +39,9 @@ import { DCF, GEN3 } from 'common/constants';
 import ArrangerAdmin from 'components/ArrangerAdmin';
 import ErrorBoundary from 'ErrorBoundary';
 import ROUTES from 'common/routes';
+import isEmpty from 'lodash/isEmpty';
+import { Spin, Icon } from 'antd';
+import { isAdminToken, validateJWT } from 'utils';
```
jberube: :+1:

src/App.js
```diff
@@ -241,52 +265,54 @@ const App = compose(
             })
           }
         />
+        <Route path={ROUTES.dashboard} exact render={props => showDashboardIfLoggedIn(props)} />
```
jberube: It's much cleaner this way!

Since `render` accept a function of `props`, you could do `<Route path={ROUTES.dashboard} exact render={showDashboardIfLoggedIn} />` instead.

src/components/Page.js
```diff
@@ -23,11 +23,17 @@ const FloatFooterPageComponentWrapper = styled('div')`
...
+  Foot = Footer,
+  Component,
+  cssFloatFooterPageComponentWrapper = ``,
+  ...props
+}) => (
   <Container>
     <Head />
     <FloatFooterPageContentWrapper>
-      <FloatFooterPageComponentWrapper>
+      <FloatFooterPageComponentWrapper css={cssFloatFooterPageComponentWrapper}>
```
jberube: This will clash with the refactor I'm doing since I'm removing the possibility of injecting css using the `css` property.

Can you use a `className` with standard css (or less, sass, etc...) or a `style` instead?

evans-g-crsj: for now, to avoid clash and all, I won't style the spinner. I'll wait untill you finish your refactor

src/App.js
evans-g-crsj: changed

src/components/Page.js
evans-g-crsj: I will avoid styling the spinner for now.

PR #2166, opened 2019-11-25
:lipstick: Override ant design styles for titles

adipaul1981:
- Changed import order in Index.js so that our styles overrides Ant Design (not the other way around) 
- Updates Style for titles (H1-H2-H3) for Ant Design


src/components/MemberSearchPage/FilterDrawer.js
```diff
@@ -30,14 +30,7 @@ class FilterDrawer extends Component {
         <div style={{ height: 50, display: 'flex', padding: '15px 7px 15px 12px' }}>
           <Title
             level={2}
-            className={'h2-title'}
             style={{
```
evans-g-crsj: why not using a conditional className instead ?

PR #2247, opened 2020-01-15
remove capitalization on interests + add basic validators (max length…

evans-g-crsj:
…) to inputs

### DIFF
--- a/src/components/UserProfile/ResearchInterests.js
+++ b/src/components/UserProfile/ResearchInterests.js
@@ -31,7 +31,7 @@ const ResearchInterest = ({ interests }) => {
           {Boolean(interestRight) && (
             <Text>
               <Icon type="check-circle" theme="filled" className={'ri-icon'} />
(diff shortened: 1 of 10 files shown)


src/components/UserProfile/ResearchInterests.js
```diff
@@ -25,14 +24,14 @@ const ResearchInterest = ({ interests }) => {
...
-            {capitalize(interestLeft)}
+            {interestLeft}
           </Text>
         </div>
         <div className={'ri-ro-text-wrapper ri-text-wrapper'}>
           {Boolean(interestRight) && (
             <Text>
               <Icon type="check-circle" theme="filled" className={'ri-icon'} />
-              {capitalize(interestRight)}
+              {interestLeft}
```
jecos: Should be interestRight

evans-g-crsj: good catch!

PR #2258, opened 2020-01-20
2251 reintroduce ga member page

evans-g-crsj:
first iteration to reintroduce ga tracking in the member page.


src/components/UserProfile/common.js
```diff
@@ -0,0 +1,35 @@
+import { trackUserInteraction, TRACKING_EVENTS } from 'services/analyticsTracking';
```
francisl: I'm not conviced by that file name. Can't understand what is it purpose.

Why not something like TrakingHandler or something like that?

PR #2304, opened 2020-02-05
fix login page flash and redirect when token is expired

francisl:


src/App.js
```diff
@@ -64,7 +64,7 @@ const userIsLoggedInButMustCompleteJoinForm = loggedInUser => {
 };
 
 const forceSelectRole = ({ loggedInUser, isLoadingUser, WrapperPage = Page, ...props }) => {
-  if (userIsRequiredToLogIn(loggedInUser)) {
+  if (userIsRequiredToLogIn(loggedInUser) && !isLoadingUser) {
     return isLoadingUser ? null : (
```
evans-g-crsj: isLoadingUser at line 67 will always be false. I think we should remove the condition.

src/services/login.js
```diff
@@ -32,6 +66,24 @@ export const facebookLogin = token =>
     },
   });
 
+export const facebookRefreshToken = async () =>
+  facebookLogin((await facebookStatus()).authResponse.accessToken);
+
+export const facebookStatus = async () => {
```
evans-g-crsj: style-wise, should we avoid mixing async-await and old school promises in the same code block?

francisl: I don't see any other option. Facebook api don't support promise, so its not possible to use await. So we have to use callback. But we need to trigger a resolve.

Do you have an idea?

src/App.js
francisl: it could be true, depending on what happen on the freactal initialization.

src/services/login.js
evans-g-crsj: Didn't know it did not support it...

src/App.js
evans-g-crsj: I meant : because of the condition the 'isLoadingUser' at line 68 will never evaluate to to true at that line since the if condition above prevents it.

francisl: fixed

PR #2307, opened 2020-02-07
Feature/2293 enable admi nusertoseeallmembers

adipaul1981:
- New filter allowing ADMIN user to see all members (public + private)
- To Have access user needs to be ADMIN
- Integrated with all other filters (Member Categories and Research Interests)
- Table header and pagination should change to display correct result.
- Tag is added for new Member Search Option Filter

image


src/components/MemberSearchPage/ADMINFilter.js
```diff
@@ -0,0 +1,137 @@
...
+  requestADMINOptionsUpdate,
+  requestCurrentPageUpdate,
+} from 'components/MemberSearchPage/actions';
+import FilterTable from 'components/MemberSearchPage/FilterTable';
+import FilterTableList from 'components/MemberSearchPage/FilterTableList';
+import { getCurrentEnd, getCurrentStart, getSelectedFilter } from './utils';
+import PropTypes from 'prop-types';
+import { ADMIN_OPTIONS } from 'common/constants';
+
+class ADMINFilter extends Component {
```
evans-g-crsj: I think AdminFilter looks better than ADMINFilter

src/components/MemberSearchPage/ADMINFilter.js
```diff
@@ -0,0 +1,137 @@
+import React, { Component } from 'react';
```
evans-g-crsj: It's seems to me that all of your filter components have a very similar structure (code duplication?) .Would it be possible to extract the repeating logic in one component?  Eg : const AdminFilter = <MyGenericFilter config={...}/> ;;; const rolesFilter = <MyGenricFilter config={...} /> , ....

src/components/MemberSearchPage/FilterTagContainer.js
```diff
@@ -4,21 +4,32 @@ import FilterTagContent from 'components/MemberSearchPage/FilterTagContent';
 
 const FilterTagContainer = ({ filters, clearTag }) => {
   return (
-    <Row className={'flex'} gutter={16} style={{ paddingTop:32, marginLeft: 0, marginRight: 0 }}>
+    <Row className={'flex'} gutter={16} style={{ paddingTop: 32, marginLeft: 0, marginRight: 0 }}>
```
evans-g-crsj: since you already use css why not extracting the inline-style into a css class?

src/components/MemberSearchPage/FilterTagContent.js
```diff
@@ -1,11 +1,16 @@
...
+import { ADMIN_OPTIONS, ROLES } from 'common/constants';
 
-const userRoleDisplayName = userRole => {
-  const role = find(ROLES, { type: userRole });
-  return role ? role.displayName : userRole;
+const userRoleDisplayName = tagName => {
+  const roleTagName = find(ROLES, { type: tagName });
+  const memberOptionTagName = ADMIN_OPTIONS[tagName];
+  return roleTagName
+    ? roleTagName.displayName
```
evans-g-crsj: see : https://eslint.org/docs/2.0.0/rules/no-nested-ternary

src/components/MemberSearchPage/MemberSearchContainer.js
```diff
@@ -77,9 +81,13 @@ class MemberSearchContainer extends Component {
       queryString,
       rolesFilter,
       interestsFilter,
+      adminOptionsFilter,
       currentPageUpdate,
     } = this.props;
-    const maxPage = count.public / membersPerPage;
+
+    const showAll = adminOptionsFilter['allMembers'] && this.props.isAdmin;
```
evans-g-crsj: adminOptionsFilter.allMembers looks better to me. + why not include isAdmin in the deconstructing object above? {otherListedProps, currentPageUpdate, isAdmin } = this.props;

src/components/MemberSearchPage/FilterTagContainer.js
adipaul1981: Its not really possible in this case...

src/components/MemberSearchPage/MemberSearchContainer.js
```diff
@@ -124,36 +135,55 @@ class MemberSearchContainer extends Component {
       membersPerPage,
       rolesFilter,
       interestsFilter,
+      adminOptionsFilter,
       updateInterestsFilter,
       updateRolesFilter,
+      updateADMINOptionsFilter,
     } = this.props;
 
+    console.log(value, 'value', type, 'Type', 'In clear tag');
```
francisl: clean

src/components/MemberSearchPage/MemberTable.js
```diff
@@ -60,6 +60,8 @@ const MemberTable = props => {
   const firstItem = props.currentPage * props.membersPerPage - props.membersPerPage + 1;
   const lastItem = props.currentPage * props.membersPerPage;
 
+  const memberShowThisPage = props.showAll ? props.count.total : props.count.public;
```
francisl: I think you have a good candidate for props destructuring, plus it will match the rest of your code style

src/common/constants.js
```diff
@@ -126,12 +130,13 @@ export const STUDY_SHORT_NAMES = [
   'Congenital Diaphragmatic Hernia',
 ];
 
+const protocol = window.location.protocol;
+const host = window.location.host;
+
```
francisl: const URI = `${protocol}://${host}`

not required, but it was more what I meant. But still better now.

PR #2331, opened 2020-02-21
Feature/2283 inactivate account

adipaul1981:
Only for ADMIN:

- Show profile status in members table (possible values: Public/Private/Deactivated):

image

in Profile Page Admin users can deactivate a profile: 
 - By deactivating a profile you also make them private
- a deactivated user (not admin) cannot make his profile public until reactivated)

image


src/components/MemberSearchPage/MemberTable.js
```diff
@@ -190,6 +220,31 @@ const MemberTable = ({
                     story={(item.highlight || {}).story || []}
                   />
                 </Col>
+                {isAdmin ? (
```
evans-g-crsj: if null is possible, I would do : isAdmin && (<Row>...</Row>)

Issue #2345, opened 2020-02-28
CohortBuilder component refactor, antd, typescript


francisl:
In the process of adding the ontologies in the cohort builder. 

Some refactoring to move toward antd and typescript.
The menu panel and modal buttons has been upgraded.

PR #2346, opened 2020-02-29
fix: #2345 upgrade antd, add full typescript support

francisl:
#2345 
While looking into the ontologies. 

I found out that there was flow types in some files that were not use it causes some issues with typescript

- Fix legacy/unused flow types
- Upgrade antd to be ready for 4.0
- Add Antd button in the cohort builder menu filters to test and also give small example for typescript

menu_filter_antd


src/components/CohortBuilder/FieldFilterContainer.js
```diff
@@ -0,0 +1,80 @@
...
+      if (node.value.length > 0 && node.value > -1) {
+        ranges++;
+      }
+    });
+    return setDisabled(
+      checkboxes.length < 1 &&
+        toggles.length < 1 &&
+        (ranges < rangeNodes.length || rangeNodes.length === 0),
+    );
+  };
```
evans-g-crsj: It is just a question, maybe it d make things worst but could this be extracted outside de FieldFilterContainer?

const hasSelectedElements = (handler, setDisableCB) => {....setDisableCB(conds)}

francisl: in another file or outside the funciton?

evans-g-crsj: outside the func if not needed elsewhere

PR #2378, opened 2020-03-27
perf(FileRepo): use cardinality to obtain stats (n of participants, n…

evans-g-crsj:
… of families, file size, n of files)

### DIFF
--- a/src/components/FileRepo/Stat.tsx
+++ b/src/components/FileRepo/Stat.tsx
@@ -4,43 +4,32 @@ import QueriesResolver, {
   Queries,
   QueriesResolverFaCCType,
 } from '../CohortBuilder/QueriesResolver';
-import {
-  familyStatTag,
-  file_stat_tag,
-  fileSizeStatTag,
-  participantStatTag,
-} from './statsConstants';
+import { Tags } from './statsConstants';
 import { WarningTwoTone } from '@ant-design/icons';
 import './FileRepo.css';
 
 interface Props {
   api: Function;
   /* helps to debug in the network console */
-  tag:
-    | typeof participantStatTag
-    | typeof familyStatTag
-    | typeof fileSizeStatTag
-    | typeof file_stat_tag;
+  tag: Tags;
   queries: Queries;
   label: string;
   icon: React.ReactElement;
 }
 
-const Stat: FunctionComponent<Props> = ({ api, tag, queries, label, icon }) => {
-  return (
-    <QueriesResolver name={`GQL_${tag}`} api={api} queries={queries}>
-      {({ isLoading, data: count, error }: QueriesResolverFaCCType) => {
-        const content = error ? <WarningTwoTone twoToneColor={'red'} /> : count;
-        return (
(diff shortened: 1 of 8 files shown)


src/components/FileRepo/Stat.tsx
```diff
@@ -0,0 +1,46 @@
...
+    | typeof familyStatTag
+    | typeof fileSizeStatTag
+    | typeof file_stat_tag;
+  queries: Queries;
+  label: string;
+  icon: React.ReactElement;
+}
+
+const Stat: FunctionComponent<Props> = ({ api, tag, queries, label, icon }) => {
+  return (
```
francisl: no need for return

src/components/FileRepo/Stat.tsx
```diff
@@ -0,0 +1,46 @@
...
+  fileSizeStatTag,
+  participantStatTag,
+} from './statsConstants';
+import { WarningTwoTone } from '@ant-design/icons';
+import './FileRepo.css';
+
+interface Props {
+  api: Function;
+  /* helps to debug in the network console */
+  tag:
```
francisl: I think an enum here will make the code cleaner/clearer https://www.typescriptlang.org/docs/handbook/enums.html

src/components/FileRepo/statsConstants.tsx
```diff
@@ -0,0 +1,4 @@
+export const file_stat_tag = 'STAT_FILE_TOTAL';
```
francisl: see above comment about enum

src/components/FileRepo/statsQueries.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const formatCount = (numOrString: number | string): string => numOrString.toLocaleString();
+
+const fromSqonValueToVariables = (content: any): any => {
+  if (!content) {
+    return;
+  }
+  return {
+    sqon: content,
+  };
```
francisl: `=> content ? { spon: content } : undefined`
?

src/components/FileRepo/statsQueries.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+export const buildStatQueriesForFileHits = (sqonContent: Object): Queries => [
+  {
+    query: queryFileStat,
+    variables: fromSqonValueToVariables(sqonContent),
+    transform: getFileStatHits,
+  },
+];
+
+const queryParticipantStat: string = `
```
francisl: no need to type, the type infered knows its a string when you assigned a value at the same time as declaring it.

valid for other assignment in this file

src/components/FileRepo/statsQueries.tsx
```diff
@@ -0,0 +1,107 @@
...
+        stats {
+          sum
+        }
+      }
+    }
+  }
+}
+`;
+
+const getFileStats = (rawData: any): string => {
```
francisl: Type inferer should work there also

PR #2395, opened 2020-04-07
Ontologies

francisl:
Screen Shot 2020-04-07 at 7 32 48 PM


src/components/CohortBuilder/Categories/Category.js
```diff
@@ -54,6 +52,21 @@ export default class Category extends React.Component {
     });
   }
 
+  onOntologyClicked = () => {
+    const isModalVisible = true;
+    this.setState({
+      isOntologyModalVisible: isModalVisible,
```
evans-g-crsj, line 58: Why not using false or true since the property name already explains the context?

src/components/CohortBuilder/Categories/Category.js
```diff
@@ -70,11 +83,11 @@ export default class Category extends React.Component {
     if (!field) return null;
 
     const { sqon, onSqonUpdate } = this.props;
-
     return (
       <Filter
         initialSqon={sqon}
         onSubmit={sqon => {
+          console.log('onsubmit : ', sqon, ' title ', title);
```
evans-g-crsj, line 90: forgot to remove

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -0,0 +1,139 @@
...
+  targetKeys: Array<string>;
+};
+
+type SelectionTreeState = {
+  treeData: TreeNode[];
+  expandedKeys: string[]
+};
+
+const getAllKeys = (data: TreeNode[], collectedKeys: string[] = [], counter = 1) => {
+  if (counter < 3) {
```
evans-g-crsj, line 22: magic number?

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -0,0 +1,139 @@
...
+      collectedKeys.push(node.key);
+      if (node.children) {
+        getAllKeys(node.children, collectedKeys, counter);
+      }
+    });
+  }
+  return collectedKeys;
+};
+
+export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeState> {
```
evans-g-crsj, line 34: Question ? In this case, it is more convenient to use a class vs hooks?

src/components/OntologyBrowser/store.test.ts
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+// import { render } from '@testing-library/react';
```
evans-g-crsj, line 2: is that comment useful?

src/components/OntologyBrowser/SelectionTree.tsx
adipaul1981: yes its the number of level to expand. it was decided to be 3. But guess it could be in a constant.

Issue #2398, opened 2020-04-08
Perf(CohortBuilder, Search All)


evans-g-crsj:
A lot of filters exist in the cohort builder in order the build a query. One of these filter is the "Search All" input that allows to search specific search.

A lot of data are loaded and the start begins even when only one character is typed. 

To relieve some pressure make sure that the search starts when 2 characters are typed.

PR #2399, opened 2020-04-08
Perf(CohortBuilder, SearchAll): #2398 search filtered data only when …

evans-g-crsj:
…a minimum number of characters are typed.

### DIFF
--- a/src/components/CohortBuilder/SearchAll/index.js
+++ b/src/components/CohortBuilder/SearchAll/index.js
@@ -84,7 +84,7 @@ const DISPLAY_MODE = {
   FILTER: 'FILTER',
 };
 
-const N_OF_CHARS_BEFORE__SEARCHING = 2;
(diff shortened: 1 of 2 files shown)


src/components/CohortBuilder/SearchAll/index.js
```diff
@@ -177,13 +179,16 @@ class SearchAll extends React.Component {
...
+    if (!canStartSearch) {
+      return null;
+    }
 
     return (
       <QueryResults
         query={debouncedQuery}
         isLoading={isLoading}
-        filteredFields={filteredFields}
+        filteredFields={getFilteredFields(debouncedQuery, fields, data[0])}
```
jecos, line 191: Where this function is defined?

evans-g-crsj: top of the file

Issue #2412, opened 2020-04-15
[Ontology Browser] Allow clearing of multiple items at once from target box


luclemo:
Currently a user can select/check multiple items in the target box, but when they click the button to remove them from the target, only one item is removed at a time. The desired behaviour: remove all checked items with a single click.

PR #2428, opened 2020-04-16
fix(ontologyBrowser): #2412 enable clearing multiple items at once

adipaul1981:


src/components/OntologyBrowser/index.tsx
```diff
@@ -111,7 +111,7 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
       });
     } else if (direction === 'left') {
       this.setState({
-        targetKeys: targetKeys.filter(key => !!moveKeys.indexOf(key)),
+        targetKeys: targetKeys.filter(key => !moveKeys.includes(key)),
```
francisl, line 114: includes is not supported in IE

adipaul1981: je suis revenu a indexOf..

PR #2430, opened 2020-04-17
Feature/2413 ontology browser improve search

adipaul1981:
- The search and expand triggers only when a minimum of 3 characters are entered
- It expands the tree and puts in bold all terms that are matched
- All other nodes with no matches are not displayed.
- If search query is less than 3 characters or empty, the tree is collapsed to its initial state (2 level expand) and all nodes are shown again.

- clear (an 'x') was added to the search bar.


src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,65 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
+      hitTreeNodes.push(treeNode.key)
+      if(result) {
+        const [ before, hit, after ] = treeNode.text.split(regex)
+        treeNode.title = <span>{before}<b>{hit}</b>{after}</span>
+      }
+    }
     return match;
   };
 
+  unhideAll = (treeNode: TreeNode) => {
```
francisl, line 129: term for tree is normal expand: `expandAll = (...`

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,65 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
+        treeData: treeData,
+        expandedKeys: getInitialKeysForExpand(treeData),
+      });
+    }
   };
 
-  searchInTree = (searchText: string, treeNode: TreeNode) => {
-    const regex = new RegExp(searchText, 'gi');
-    const text = treeNode.title as string;
+  searchInTree = (searchText: string, treeNode: TreeNode, hitTreeNodes: string[] = []) => {
```
jecos, line 102: That could be super nice to have test for this function

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,65 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
+    let regex = new RegExp('\\b(\\w*' + searchText + '\\w*)\\b', 'gi')
```
jecos, line 103: Just by curiosity why let instead of const. Why semicolon disapeared

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,65 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
       });
       match = matchChild || match;
     }
-
     treeNode.hidden = !match;
+
+    if(!treeNode.hidden) {
+      hitTreeNodes.push(treeNode.key)
+      if(result) {
+        const [ before, hit, after ] = treeNode.text.split(regex)
```
jecos, line 122: Is it ok if there is no before or after?

src/components/OntologyBrowser/SelectionTree.tsx
adipaul1981: Actually its a mistake. In this case const might be better. i don;t want `regex` to be re-assignable here. I've changed it for const. good catch.

src/components/OntologyBrowser/SelectionTree.tsx
adipaul1981: No, its not ok. But it should never happen. There is a check before hand `if(result) {...}`, therefore there is a match with the regex. If so, before and after cannott be `undefined`. One might be empty strings if the regex match is at the beginning/end or both empty is the full term is matched by the regex.

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,66 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
     selectedKeys.indexOf(eventKey.toString()) !== -1;
 
   onChange = (e: React.ChangeEvent<HTMLInputElement>, treeData: TreeNode[]) => {
-    treeData.forEach(node => this.searchInTree(e.target.value, node));
-    this.setState({ treeData });
+    const hits: string[] = []
+
+    if(e.target.value.length >= MIN_SEARCH_TEXT_LENGTH){
```
francisl, line 84: I don't know what you think, but I think because the if/else don't returned, seting the state at the end make the code more consise

```javascript
    const hits: string[] = []
    let newExpandedKeys: string[] = []

    if(e.target.value.length >= MIN_SEARCH_TEXT_LENGTH){
      treeData
        .forEach(node => this.searchInTree(e.target.value, node, hits))
      newExpandedKeys = hits;
    } else {
      treeData.forEach(node => this.showAll(node))
      newExpandedKeys = getInitialKeysForExpand(treeData)
    }

    this.setState({
      treeData: treeData,
      expandedKeys: getInitialKeysForExpand(treeData),
    });
```
its up to you

src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -78,30 +79,66 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
+  showAll = (treeNode: TreeNode) => {
+    treeNode.hidden = false
+    treeNode.title = treeNode.text
+
+    if (treeNode.children.length > 0) {
+      treeNode.children.forEach((child: TreeNode) => {
+          this.showAll(child)
+      });
+    }
+    return null;
```
francisl, line 139: isn't undefined fine here?

src/components/OntologyBrowser/Test/SelectionTree.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+describe('In SelectionTree', () => {
+  const checkedKeys: string[] = [];
+  const targetKeys: string[] = [];
+  const onItemSelect = jest.fn();
+  const dataSource: TreeNode[] = treeData
+
+  beforeAll(() => {
+  });
+
+  it('should renders without crashing', () => {
```
francisl, line 19: 'should renders selection tree' better to have positive structure.

Also, you dont expect anything here, could fail silently. It need at least one expect

Issue #2434, opened 2020-04-18
Ontology add loading indicator in modal


francisl:
data can be long to fetch, need a loading indicator

PR #2439, opened 2020-04-21
Feature(Ontlogy Browser Modal): #2434 add a spinner when data is load…

evans-g-crsj:
…ing, show error when data fetching is broken + use 'updateData()' in componentDidMount + use styling consistency (const {...} = this.props everywhere vs a mix of that and this.props.myprop). Todo: add testing

### DIFF
--- a/src/components/OntologyBrowser/OntologyModal.test.tsx
+++ b/src/components/OntologyBrowser/OntologyModal.test.tsx
@@ -1,3 +1,64 @@
-describe('OntologyModal', () => {
-  test.todo('add test');
+import React from 'react';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+import OntologyModal from './index';
+import { configure, mount, ReactWrapper } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import { Spinner } from '../../uikit/Spinner';
+
+configure({ adapter: new Adapter() });
+
+const mockStore = configureStore();
+const store = mockStore();
+
+describe('Ontology Modal', () => {
+  const props = {
+    isVisible: true,
+    initialSqon: { op: 'and', content: [] },
+    title: 'Clinical',
+    selectedField: 'observed_phenotype.name',
+  };
+
+  let wrapper: ReactWrapper;
+  beforeEach(() => {
+    wrapper = mount(
+      <Provider store={store}>
+        <OntologyModal {...props} />
+      </Provider>,
+    );
+  });
+
+  afterAll(() => {
+    wrapper.unmount();
+  });
+
+  it('should render Ontology Modal', () => {
+    expect(wrapper.length).toEqual(1);
+  });
+
+  it('should display a Spinner when Ontology Modal is in a loading state', () => {
+    const foundOntologyModal = wrapper.find('OntologyModal');
+    foundOntologyModal.setState({
+      selectedKeys: [],
+      targetKeys: [],
+      isLoading: true,
+      error: null,
+    });
+
+    expect(foundOntologyModal.state('isLoading')).toEqual(true);
+    expect(foundOntologyModal.find(Spinner).exists());
+  });
+
+  it('should not display a Spinner when Ontology Modal is not in a loading state', () => {
+    const foundOntologyModal = wrapper.find('OntologyModal');
+    foundOntologyModal.setState({
+      selectedKeys: [],
+      targetKeys: [],
+      isLoading: false,
+      error: null,
+    });
+
+    expect(foundOntologyModal.state('isLoading')).toEqual(false);
+    expect(!foundOntologyModal.find(Spinner).exists());
+  });
 });
--- a/src/components/OntologyBrowser/OntologyModal.test.tsx
+++ b/src/components/OntologyBrowser/OntologyModal.test.tsx
@@ -1,0 +1,3 @@
+describe('OntologyModal', () => {
+  test.todo('add test');
+});
--- a/src/components/OntologyBrowser/index.tsx
+++ b/src/components/OntologyBrowser/index.tsx
@@ -1,11 +1,12 @@
 import * as React from 'react';
 import { connect } from 'react-redux';
-import { Modal, Transfer, Spin, Empty, Result } from 'antd';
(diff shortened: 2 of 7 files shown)


src/components/OntologyBrowser/index.tsx
```diff
@@ -149,43 +197,57 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
...
+        ) : (
+          <Transfer
+            dataSource={dataSource}
+            targetKeys={targetKeys}
+            onChange={this.onChange}
+            render={(item: TransferItem): RenderResult => item.title || item.key}
+            disabled={false}
+            showSelectAll={false}
+            locale={{
+              notFoundContent: generateMsgWhenHasDataButNoSelection(hasData, isLoading) || (
```
francisl, line 228: I think it would be clearer here if you has a function that return the boolean value of your business logic, not something that return string sometime or null and then check.

e.g. 
```javascript
isLoadingOrEmtpy = () => (this.state.isLoading || dataSource.length === 0)

...
         notFoundContent: this.isLoadingOrEmtpy() ? 
              'Select items from the panel on the left in order to add them to your query' :
              <Empty image={Empty.PRESENTED_IMAGE_SIMPLE} />
```

src/components/OntologyBrowser/index.tsx
```diff
@@ -54,17 +56,48 @@ const updateSqons = (initialSqon: Sqon, value: string[]) => {
   return [initialSqon];
 };
 
+const generateMsgWhenHasDataButNoSelection = (hasData: boolean, isLoading: boolean) => {
```
francisl, line 59: 👇

src/components/OntologyBrowser/index.tsx
```diff
@@ -149,43 +197,57 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
...
-          render={(item: TransferItem): RenderResult => item.title || item.key}
-          disabled={false}
-          showSelectAll={false}
-        >
-          {({ direction, onItemSelect, selectedKeys }) => {
-            if (direction === 'left') {
-              const checkedKeys = [...selectedKeys, ...this.state.targetKeys];
-              if (this.state.treeSource) {
+        {hasError ? (
+          displayError()
```
francisl, line 218: I don't see value in this function jump, it doesn't make the code easier to understand

src/components/OntologyBrowser/index.tsx
```diff
@@ -54,17 +56,48 @@ const updateSqons = (initialSqon: Sqon, value: string[]) => {
...
 
+const generateMsgWhenHasDataButNoSelection = (hasData: boolean, isLoading: boolean) => {
+  if (isLoading || !hasData) {
+    return null;
+  }
+  return 'Select items from the panel on the left in order to add them to your query';
+};
+
+const displaySpinner = () => (
+  <div
```
francisl, line 67: Style shoud go in a css file, plus, I think this should go in a separate component. We should have something like uikit but with antd for pre-baked KF components

src/components/OntologyBrowser/index.tsx
```diff
@@ -92,15 +128,21 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
     return Object.values(results);
   };
 
+  closeAndCleanModal = () => {
+    const { onCloseModal } = this.props;
+    this.setState({ error: null });
+    onCloseModal();
```
francisl, line 134: I don't see any value to desconstruct props here

```javascript
closeAndCleanModal = () => {
    this.setState({ error: null });
    this.props.onCloseModal();
}

```

src/components/OntologyBrowser/index.tsx
```diff
@@ -123,16 +165,22 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
     });
   };
 
-  updateData = () => {
-    this.ontologyStore.fetch(this.props.initialSqon).then(() => {
+  updateData = async () => {
+    this.setState({ isLoading: true });
+    const { initialSqon } = this.props;
+    try {
+      await this.ontologyStore.fetch(initialSqon);
```
francisl, line 172: same thing here, if the props is only use one time on a short line, maybe desconstructing it add no or little value. What do you think?

src/components/OntologyBrowser/index.tsx
```diff
@@ -123,16 +165,22 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
...
       this.transfertDataSource = [];
       this.flattenDataSource(this.ontologyStore.tree as TransferItem[]);
       const newTargetKeys = this.getKeysFromSqon();
       this.setState({
         treeSource: this.ontologyStore.tree,
         targetKeys: newTargetKeys,
+        isLoading: false,
       });
-    });
+    } catch (e) {
```
francisl, line 181: Maybe it's better to just wrap the try/catch in what might cause the exception. It's better pratice to scope down so it clear in the future what might throw

src/components/OntologyBrowser/index.tsx
evans-g-crsj: What you say makes sense but I kept the same 'style' (deconstructing) everywhere in the file. But I'll do what you suggest.

src/components/OntologyBrowser/index.tsx
evans-g-crsj: Good point.

src/components/OntologyBrowser/index.tsx
evans-g-crsj: Good point.

src/components/OntologyBrowser/index.tsx
evans-g-crsj: Good point.

src/components/OntologyBrowser/index.tsx
evans-g-crsj: Very good point

src/components/OntologyBrowser/OntologyModal.test.tsx
```diff
@@ -0,0 +1,3 @@
+describe('OntologyModal', () => {
```
francisl, line 1: need fix

Issue #2425, opened 2020-04-16
SQON BUILDER - Participant count not updated un adding observed HPO term


adipaul1981:
![image.png](https://images.zenhubusercontent.com/5db1b4ee9f4bf300014d8774/854f7fb3-28d6-48d7-831b-2b47d9cc4ed9)

![image.png](https://images.zenhubusercontent.com/5db1b4ee9f4bf300014d8774/fcb51ed3-8e4d-400f-8afd-d998b3964902)

PR #2452, opened 2020-04-22
fix(SqonBuilder): #2425 display correct participants count

evans-g-crsj:


src/components/CohortBuilder/Categories/Category.js
```diff
@@ -115,7 +115,7 @@ export default class Category extends React.Component {
             isVisible={isOntologyModalVisible}
             onCloseModal={this.onOntologyModalClose}
             initialSqon={sqon}
-            onSqonUpdate={sqon => {
+            onSqonUpdateCB={sqon => {
```
jecos, line 118: Why do we need to change the name?

PR #2528, opened 2020-05-19
2202 improve download report

evans-g-crsj:
(CohortBuilder -> Table View) + File Repo
screenshotDownloadBtnFileRepo


src/components/FileRepo/DownloadButton.tsx
```diff
@@ -0,0 +1,216 @@
...
+};
+
+type ColumnStateRenderProps = {
+  state: StateProp;
+  loading: boolean;
+};
+
+const trackDownload = async (label: string) =>
+  // @ts-ignore
+  await trackUserInteraction({
```
francisl, line 68: can you provide with more information on why this is not solvable here? `@ts-ignore` I know it's sometime a necessary evil, but we should justify it a little at least.

src/components/FileRepo/DownloadButton.tsx
```diff
@@ -0,0 +1,216 @@
...
+    columns,
+    isFileRepo: isFileRepo,
+  };
+  return clinicalDataParticipants(downloadConfig);
+};
+
+const familyDownloader = async ({ api, sqon, columns }: DownloaderParams) => {
+  await trackDownload('Clinical (Participant and family)');
+  const isFileRepo = true;
+  const { familyMemberIds, participantIds } = await familyMemberAndParticipantIds({
```
francisl, line 99: how was the api handle before, don't we have a reducer/actions for this business logic?

francisl: What I mean is that business logic about fetching data should not be bind the the UI component, but normally to a action/reducer/store

Was it in the same file before?

src/components/FileRepo/DownloadButton.tsx
francisl: is it easily fixable, need major refactoring?

src/components/FileRepo/DownloadButton.tsx
evans-g-crsj: Good point. Yes, it was in the same file. I'll keep that way for now. But, There s another ticket coming up where I might have to change these actions. I'll take that opportunity to unbind that logic

src/components/FileRepo/DownloadButton.tsx
evans-g-crsj: I can not completely remove the 'ts-ignore' but I improved the situation based on your comment. It will be fixed when we refactor the tracking function from plain javascript into typescript.

francisl:
LGTM

PR #2548, opened 2020-05-28
Feature/2516 ontology browser selected option path

adipaul1981:
image


src/components/OntologyBrowser/SelectionTree.tsx
```diff
@@ -146,25 +145,52 @@ export class SelectionTree extends Component<SelectionTreeProps, SelectionTreeSt
...
+  onExpand = (expand: (string | number)[]) => {
     this.setState({
       expandedKeys: expand.map((v) => v.toString()),
     });
   };
 
+  getCheckedKeys = (
+    key: string | number,
+    dataSource = this.props.dataSource,
+    accu: { check: string[]; halfcheck: string[] } = {
```
francisl, line 157: imo, i'll keep a name like checkKeys or something like that, It's more a javascripty than scalaist (if I can say that)

Issue #2552, opened 2020-05-29
Observed Phenotype displays error when using 3 queries at a time in the Cohort Builder


denis-crsj:
**Bug description**
When doing a 3rd query in the cohort builder that combines the results of the 2 first queries, the Observed Phenotype Browser shows an error when trying to open it.

**Steps To Reproduce**
1. Go to the Cohort Builder
2. Create a query with any filter
3. Create a second query with any filter again
4. Create a 3rd query combining the first 2 queries with the "AND" operator
5. Go into Clinical and try to open the Observed Phenotype Browser

**Expected behavior**
The browser should open normally and not display any errors

**Screenshots**
![image](https://user-images.githubusercontent.com/64538118/83278631-e448bc00-a1a1-11ea-800f-897474c32194.png)

![image.png](https://images.zenhubusercontent.com/5ea980d6189dd52c91ddddfa/1e794f0d-62a4-409d-bbae-3ace63f79b16)

**Desktop (please complete the following information):**
 - OS: macOS Catalina 10.15.4
 - Browser: Chrome
 - Version 81.0.4044.129 (Official Build) (64-bit)

**Additional context**
Bug mentioned by Vincent Ferretti

denis-crsj:
Issue resolved. Tested in Firefox and Chrome. Moving the task to Done.

PR #2554, opened 2020-06-01
fix(ontologyBrowser): #2552 enable combined queries

adipaul1981:
image
image


src/components/OntologyBrowser/index.tsx
```diff
@@ -78,23 +82,28 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
...
+      };
+      this.ontologyStore.tree.forEach((treeNode) => {
+        findTreeKey(treeNode);
       });
-    };
-    this.ontologyStore.tree.forEach((treeNode) => {
-      findTreeKey(treeNode);
-    });
+    }
     return Object.values(results);
```
francisl, line 107: why not return that in the condition and remove the big indented block?

PR #2585, opened 2020-06-27
fix(file repository): add utils to transform sqon that are file centr…

evans-g-crsj:
…ic to participant centric + add better testing for reports + add string type to value in sqon filter

### DIFF
--- a/src/store/Test/sqonUtils.test.tsx
+++ b/src/store/Test/sqonUtils.test.tsx
@@ -11,7 +11,7 @@ import {
 
 describe('SqonUtils', () => {
   describe('removeSaveSetFilters', () => {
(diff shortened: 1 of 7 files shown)


src/store/Test/sqonUtils.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    });
+
+    it('should keep sqon filter intact when no save set', async () => {
+      const sqonFilters = fileCentricSqonWithoutSaveSet.content;
+      expect(removeSaveSetFilters(sqonFilters)).toEqual(sqonFilters);
+    });
+  });
+
+  describe('extractSaveSetIdsFromSqon', () => {
+    it('should extract saveSet id when needed', async () => {
```
adipaul1981, line 35: Need to be async here?

Issue #2556, opened 2020-06-03
Ontology Browser: redundant display for terms on multiple paths


vferretti:
**Bug description**
the same term can be displayed multiple time on the right panel if it belongs to multiple path
**Steps To Reproduce**
1. Go to the ontology browser
2 search for Abnormal lower limb bone morphology (HP:0040069)
3 select twice this term from different path
4 you see the term has been added more than once on the right

**Expected behavior**
We should only see once the same search criteria

![image](https://user-images.githubusercontent.com/1181124/83665301-3b330480-a599-11ea-93af-578fd050f73c.png)

### Sub-Tasks
- [ ] Sub-task 1
- [ ] Sub-task 2
- [ ] Sub-task 3
- [ ] Sub-task 4

### Testing Strategy
See the steps to reproduce above as well as the spteps to reproduce. Anybody with access to the Bastion will be able to test.

denis-crsj:
Tested using Abnormal lower limb bone morphology (HP:0040069). Once it has been added, it cannot be added again and the box is uncheckable. Moving to Done.

PR #2601, opened 2020-07-01
fix: #2556 bug

adipaul1981:


src/components/OntologyBrowser/store.ts
```diff
@@ -26,24 +26,14 @@ const dotToUnderscore = (str: string) => str.replace('.', '__');
...
-    if (match) {
-      const term = match.pop();
+  let allSelectedAndChecked = {};
 
-      if (term) {
-        const pattern = termPattern(term);
-        updatedTargetKeys = updatedTargetKeys.filter((t) => !pattern.test(t));
-      }
-    }
+  selectedKeys.concat(targetKeys).forEach((t) => {
```
evans-g-crsj, line 32: why not using a map vs forEach so that you can avoid using let?

PR #2611, opened 2020-07-06
2566 virtual study delete confirmation modal

evans-g-crsj:


src/components/UserDashboard/SavedQueries/index.js
```diff
@@ -64,7 +66,7 @@ class SavedQueries extends React.Component {
     this.props.fetchVirtualStudiesCollection(this.props.loggedInUser.egoId);
   }
 
-  deleteVirtualStudy(vs) {
+  deleteVirtualStudy = (vs) => {
```
francisl, line 69: here you can omit the`{}` and return the actions return value. That way, if the caller want to do something with it, it will be available

PR #2636, opened 2020-08-06
Feat/2581 prevent saving save sets with same name

adipaul1981:


src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -41,114 +71,131 @@ export const validateNameSetInput = (value: string): NameSetValidator => {
...
+      <Form form={form} name={FORM_NAME} onFinish={onFinish}>
+        <Form.Item
+          label="Name"
+          name="nameSet"
+          hasFeedback
+          validateStatus={displayHelp || tagNameConflict ? 'error' : 'success'}
+          help={
+            displayHelp ? displayHelpMessage : 'Letters, numbers, hyphens (-), and underscores (_)'
+          }
+          initialValue={'Save_Set_1'}
```
evans-g-crsj, line 166: I think it would be better if the initialValue is set in the <Form/>

src/services/sets.tsx
```diff
@@ -38,5 +38,40 @@ const getIdsFromSaveSetId = async (rawId: string) => {
...
+          },
+          {
+            op: 'in',
+            content: { field: 'userId', value: [userId] },
+          },
+        ],
+      },
+    },
+  });
+  return response?.data?.sets?.aggregations?.size?.stats?.count || 0;
```
evans-g-crsj, line 70: why all these checks? Wouldn't it be better to throw an error is something that is not expected occurs?

src/store/actionCreators/saveSets.ts
```diff
@@ -0,0 +1,62 @@
+import { ThunkAction } from 'redux-thunk';
+import {
+  FAILURE_CREATE,
+  RE_INITIALIZE_STATE,
+  SaveSetParams,
+  SaveSetsActionTypes,
+  TAG_NAME_CONFLICT,
+  TOGGLE_PENDING_CREATE,
+} from '../saveSetTypes';
+import { saveSetCountForTag } from '../../services/sets';
```
evans-g-crsj, line 10: remove '../..' if possible

src/store/actionCreators/saveSets.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (tagIsUnique) {
+      await saveSet({
+        type: 'participant',
+        path: 'kf_id',
+        sqon,
+        userId: userId,
+        api: graphql(api),
+        tag: tag,
+      });
+      onSuccess();
```
evans-g-crsj, line 33: I would check if the callback is available. If that is to be reused elsewhere nothing says we would need to use a callback - it's optional

src/services/sets.tsx
```diff
@@ -38,5 +38,45 @@ const getIdsFromSaveSetId = async (rawId: string) => {
...
+          {
+            op: 'in',
+            content: { field: 'userId', value: [userId] },
+          },
+        ],
+      },
+    },
+  });
+
+  if (response?.data || undefined) {
```
evans-g-crsj, line 71: no

src/store/actionCreators/saveSets.ts
```diff
@@ -0,0 +1,86 @@
...
+  try {
+    await saveSet({
+      type: 'participant',
+      path: 'kf_id',
+      sqon,
+      userId: userId,
+      api: graphql(api),
+      tag: tag,
+    });
+    if (onSuccess && typeof onSuccess === 'function') {
```
evans-g-crsj, line 32: no need for typeof

PR #2644, opened 2020-08-07
feature(login, join): show terms and conditions when logging in.

evans-g-crsj:


src/components/Login/roleForm.css
```diff
@@ -0,0 +1,31 @@
+.text-align-left {
+    text-align: left;
+}
+
+.section-header-spacing {
+    color: #2b388f;
```
francisl, line 6: why this is not part of the theme?

src/components/Login/termsConditions.css
```diff
@@ -0,0 +1,32 @@
+.footer {
+  text-align: right;
+  height: 52px;
+  border-top: 1px solid #f0f0f0;
```
francisl, line 4: ☝️

src/App.js
```diff
@@ -21,7 +20,7 @@ import SideImagePage from 'components/SideImagePage';
 import Page from 'components/Page';
 import { FixedFooterPage } from 'components/Page';
 import ContextProvider from 'components/ContextProvider';
-import Error from 'components/Error';
+import Error from './components/Error';
```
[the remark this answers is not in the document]
evans-g-crsj: Good point. I did not notice. Thanks Intellij

src/components/Login/roleForm.css
evans-g-crsj: It's specific to one form but it's a good idea. I'll post-pone when the ticket for the theme will be done.

Issue #2581, opened 2020-06-26
Prevent user from saving multiple sets with the same name


denis-crsj:
### Feature

Each Saved Sets names should be unique. The user shouldn't be able to reuse the same name multiple times as it may become confusing for the user.

### Acceptance Criteria

**Scenario 1** Use has no saved set yet
  - **Given:** The user is in the save modal
  - **When:** The user types the name
  - **Then:** No error message is displayed since the user does not have any saved sets yet
    - **And:**The user clicks on Save to confirm and close the modal box

**Scenario 2** User has at least 1 saved set already
  - **Given:** The user is in the Save modal
  - **When:** The user types the same name of one of his sets
  - **Then:** An error message is displayed since the name already exists
    - **And:**The Save button is greyed out preventing the user to click it and close the modal box

### Sub-Tasks
_List of what needs to be done  on the technical side._
- [ ] Sub-task ...

### Testing Strategy
Anyone with access to the Bastion and QA site can test. No material needed.

### Defects
_List of defects/bugs found during QA._
- [ ] Defect ...

### Resources

![Screen Shot 2020-06-26 at 9.02.59 AM.png](https://images.zenhubusercontent.com/5ea980d6189dd52c91ddddfa/bcaebb7b-079b-4255-b25e-1c0ee8f07cc3)

PR #2647, opened 2020-08-11
fix(SaveSets): #2581 save sets name conflict qa review

adipaul1981:


src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -111,7 +107,7 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
   const maxNumOfCharsToDisplay = MAX_LENGTH_NAME + 1;
 
   useEffect(() => {
-    if (error) {
+    if (error && !isSaveSetNameConflictError(error)) {
```
evans-g-crsj, line 110: I think the first check if unnecessary: error && !isSaveSetNameConflictError(error) => !isSaveSetNameConflictError(error)

But make sure it's ok

src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -155,24 +151,18 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
...
             () => ({
               validator: (_, value) => {
                 if (hasError) {
                   reInitializeState();
                 }
-                setDisplayHelp(false);
-                setDisplayHelpMessage('');
                 const { msg, err } = validateNameSetInput(value);
-                isSaveButtonDisabled();
                 setHasError(err);
```
evans-g-crsj, line 163: why not set the error only when there is an error? Couldn't we call 'setHasError' in the if (err) block?

src/store/saveSetTypes.tsx
```diff
@@ -39,3 +39,12 @@ export type SaveSetParams = {
   onSuccess: Function;
   onNameConflict: Function;
 };
+
+export class SaveSetNameConflictError extends Error {
+  constructor(message: string | undefined) {
```
evans-g-crsj, line 44: why undefined?

PR #2651, opened 2020-08-12
fix: make sure that user stays on same route when he manually refresh…

evans-g-crsj:
…es a page + remove unneeded code.

### DIFF
--- a/src/stateProviders/provideLoggedInUser.js
+++ b/src/stateProviders/provideLoggedInUser.js
@@ -58,11 +58,14 @@ export default provideState({
             api,
             onFinish: (user) => {
               const showJoin = !hasUserRole(user);
+              const routeIsNotAvailableWhenLoggedIn = ['/join', '/orcid'].includes(
+                window.location.pathname,
+              );
               if (showJoin) {
                 history.push(ROUTES.join);
               } else if (!user.acceptedTerms) {
                 history.push(ROUTES.termsConditions);
-              } else if (['/', '/join', '/orcid'].includes(window.location.pathname)) {
+              } else if (routeIsNotAvailableWhenLoggedIn) {
                 history.push(ROUTES.dashboard);
               }
             },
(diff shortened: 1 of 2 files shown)


src/stateProviders/provideLoggedInUser.js
```diff
@@ -127,24 +118,7 @@ export default provideState({
...
-              ? without(fields, 'institution', 'jobTitle')
-              : fields;
-          const profile = pick(omit(user, 'sets'), userRoleProfileFields);
-          const filledFields = Object.values(profile || {}).filter(
-            (v) => (isArray(v) && v.length) || (!isArray(v) && v),
-          );
-          const percentageFilled = filledFields.length / userRoleProfileFields.length;
-          if (state.loggedInUser && state.percentageFilled < 1 && percentageFilled >= 1) {
-            trackUserInteraction({
-              category: TRACKING_EVENTS.categories.user.profile,
```
francisl, line 143: its ok to remove tracking?

src/stateProviders/provideLoggedInUser.js
```diff
@@ -71,8 +62,8 @@ export default provideState({
                 history.push(ROUTES.join);
               } else if (!user.acceptedTerms) {
                 history.push(ROUTES.termsConditions);
-              } else {
-                history.push(ROUTES.login);
+              } else if (['/', '/join', '/orcid'].includes(window.location.pathname)) {
```
francisl, line 65: why should it go to dashboard if you hit /join?

evans-g-crsj: it's legacy code (not visible in the left panel for I removed it). I think it means that if you type in /join in the url but you are already logged in then go to dashboard. These pages are hidden if logged in (except for '/')

src/stateProviders/provideLoggedInUser.js
evans-g-crsj: In this case, yes. The concept of 'at what % the profile is completed' doesn t exist anymore.

PR #2673, opened 2020-08-20
Feat/2573 saveset default name

adipaul1981:


src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -55,6 +57,19 @@ type PropsFromRedux = ConnectedProps<typeof connector>;
...
+export const extractTagNumbers = (userSets: [{ node: UserSaveSets }]) => {
+  const regExp = /saved_set_([0-9]+)/gi;
+
+  return userSets.reduce((acc: number[], s: { node: UserSaveSets }) => {
+    if (s.node.tag.match(regExp)) {
+      //cannot be null if match
+      // @ts-ignore
+      return [...acc, Number(regExp.exec(s.node.tag)[1])];
+    }
+    return acc;
```
francisl, line 69: the null error is call the 'billion dollar mistake', so I think we should be always explicit about it.

is the match necessary, wont be a exec and checking the result also be fine?

```
    const match = regExp.exec(s.node.tag);

    if (match !== null && match.length > 1) {
      return [...acc, Number(match[1])];
    }
```

Issue #2637, opened 2020-08-07
Adjust HPO Sunburst arc proportions based on new algorithm


denis-crsj:
### Feature

We need to adjust the sunburst based on the new algorithm that was created to calculate the arc proportions properly.

### Acceptance Criteria
- The proportions of the arc should be calculated based on the new algorithm
- The Sunburst needs to be visible in Summary view on the Cohort Builder only

### Sub-Tasks
_List of what needs to be done  on the technical side._
- [ ] Sub-task 1

### Testing Strategy
_What do we need to test? Who can test it? How do we test it? What do we need to test it?_

### Defects
_List of defects/bugs found during QA._
- [ ] Defect 1

### Resources

Sunburst specs: https://www.notion.so/ferlab/KF-Ontology-Sunburst-212f8110b7794cbca44198a1ab7f993b
Algorithm doc: https://www.notion.so/ferlab/KF-Ontology-Sunburst-212f8110b7794cbca44198a1ab7f993b#5c3befd70dc548e1abd3eb225eedb94b

PR #2678, opened 2020-08-24
feat(ontology): #2637 new sunburst algorith

francisl:
Screen Shot 2020-08-21 at 5 50 57 PM
Screen Shot 2020-08-21 at 5 51 24 PM
Screen Shot 2020-08-21 at 5 51 41 PM


src/components/OntologyBrowser/Model.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(data: PhenotypeSource[], field: string) {
+    this.phenotypes = data;
+    this.tree = this.generateTree(field);
+  }
+
+  private getChildrenValue = (childrenNodes: TreeNode[], sourceValue: number) => {
+    const value = childrenNodes.length
+      ? childrenNodes.reduce((accumulator, n) => accumulator + (n.valueText || 0), 0)
+      : sourceValue || 0;
+    return value;
```
evans-g-crsj, line 33: I would return the whole expression directly

src/components/OntologyBrowser/Model.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  createNodeFromSource = (
+    source: PhenotypeSource,
+    exactTagField: string,
+    children: TreeNode[] = [],
+    depth: number = 0,
+    parentKey?: string,
+  ): TreeNode => {
+    const value = this.getChildrenValue(children, source.doc_count);
+    // children.length !== 0 ? this.getChildrenValue(children, source.doc_count) : source.doc_count;
```
evans-g-crsj, line 82: is comment useful?

src/components/OntologyBrowser/store.ts
```diff
@@ -212,6 +123,32 @@ export class PhenotypeStore {
...
+      if (result) {
+        break;
+      }
+    }
+    return result;
+  };
+
+  getChildrenKeys = (node: TreeNode, root = false) => {
+    let nKeys: string[] = [];
+    node.children.forEach((i) => {
```
evans-g-crsj, line 143: why not using map?

src/components/OntologyBrowser/store.ts
```diff
@@ -212,6 +123,32 @@ export class PhenotypeStore {
...
+      nKeys = nKeys.concat(this.getChildrenKeys(i));
+    });
+    if (!root) {
+      nKeys.push(node.key);
```
evans-g-crsj, line 147: is using a more functional style (like using map) I would do this [...nKeys, node.key].But it s just a matter of taste

src/components/UI/Charts/Sunburst/sunburst-d3.jsx
```diff
@@ -18,20 +18,11 @@ const sunburstD3 = (ref, data, config, formatters) => {
...
-    if (d.length === 0) return 0;
-    d.reduce((accumulator, value) => accumulator + value.results + countResults(value.children));
-  };
-
   const partition = (data) => {
     const root = d3
       .hierarchy(data)
-      .sum((d) => d.results + countResults(d.children))
-      .sort((a, b) => b.results - a.results);
+      .sum((d) => d.value)
```
evans-g-crsj, line 24: out of curiosity, sum gives an array? I would have thought that .sum gives a scalar. Therefore, sorting would make no sense.

src/components/OntologyBrowser/store.ts
francisl: map wont work, but flapmap can. I'll run some benchmark to see what is optimal

src/components/OntologyBrowser/store.ts
francisl: this does not address that the mutation/concatenation is conditional

src/components/UI/Charts/Sunburst/sunburst-d3.jsx
francisl: sum is for the value of the element, sort is how to place them on the arc

PR #2688, opened 2020-09-04
Feat/2583 edit save set tag

adipaul1981:


src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -107,14 +128,46 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
...
-      tag: nameSet,
-      userId: user.egoId,
-      api: api,
-      sqon: sqon,
-      onSuccess: onSuccessCb,
-      onNameConflict: onNameConflictCb,
-    });
+    switch (saveSetActionType) {
+      case SaveSetActionsTypes.EDIT:
+        if (setToRename) {
```
evans-g-crsj, line 133: I think you can merge this condition with the switch condition => case (SaveSetActionsTypes.EDIT && setToRename)

it would remove an indentation depth

src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -182,13 +235,17 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
         </Form.Item>,
       ]}
     >
-      {loadingDefaultTagName ? (
+      {SaveSetActionsTypes.CREATE === saveSetActionType && loadingDefaultTagName ? (
```
evans-g-crsj, line 238: Are you sure it's not a || vs &&

src/store/actionCreators/saveSets.ts
```diff
@@ -91,14 +130,17 @@ export const getUserSaveSets = (
...
+  const { userId, setIds, onFail } = payload;
+
   dispatch(isDeletingSaveSets(true));
+
   try {
-    await deleteSaveSet(userId, setIds);
+    const result = await deleteSaveSet(userId, setIds);
 
-    dispatch(removeUserSavedSets(setIds));
+    if (result && result > 0) dispatch(removeUserSavedSets(setIds));
```
evans-g-crsj, line 142: I would use curly braces since not using them can create bugs are to detect

evans-g-crsj (changes requested):

src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
adipaul1981: i dont think that would work.  i can change it to this. 
case SaveSetActionsTypes.EDIT:
        if (!setToRename) {
          break;
        }
        await onEditSet({

src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
adipaul1981: yes its ok. there can only be a spin in the case of a create of Save set

evans-g-crsj (approved):

PR #2689, opened 2020-09-04
fix style issues reported by QA

evans-g-crsj:


francisl (approved):
is  `wrapperVerticalBarChart` still in use?

evans-g-crsj:
> is `wrapperVerticalBarChart` still in use?

yes, just not visible in the files differences

PR #2691, opened 2020-09-10
refactor: #2633 refactor summary table

francisl:
#2633 refactor summary table

Screen Shot 2020-09-10 at 3 15 57 AM

* Missing action bar refactor


src/components/CohortBuilder/Results.js
```diff
@@ -166,92 +167,110 @@ const Results = ({
...
+          {showDetailsHeader()}
+          {isLoading ? (
+            <div className={'cb-summary-is-loading'}>
+              <Spin size={'small'} />
+            </div>
+          ) : (
+            <div className="cb-summary">
+              <div className="cb-summary-entity">
+                <h3 className="cb-sub-heading">
+                  <DemographicIcon width="14px" height="17px" />
```
evans-g-crsj, line 196: do we have antd equivalent for icons and images?

src/components/CohortBuilder/Summary/Cards/SummaryCard.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { Card as AntdCard } from 'antd';
+import './SummaryCard.css';
+
+type CardProps = {
+  title?: string;
+  loading?: boolean;
+  children: ReactElement[] | any | Function;
+};
+
+const Card = ({ title, loading, children, ...props }: CardProps) => {
```
evans-g-crsj, line 11: can we make ...props explicit or at least as explicit as possible? By explicit I mean to replace ...props with all the values it represents

src/components/CohortBuilder/Summary/Cards/SummaryCard.tsx
```diff
@@ -0,0 +1,28 @@
+import React, { ReactElement, useRef, RefObject } from 'react';
+import { Card as AntdCard } from 'antd';
+import './SummaryCard.css';
+
+type CardProps = {
+  title?: string;
+  loading?: boolean;
+  children: ReactElement[] | any | Function;
```
evans-g-crsj, line 8: or any?

evans-g-crsj (approved):

src/components/CohortBuilder/Results.js
francisl: shuffled code

src/components/CohortBuilder/Summary/Cards/SummaryCard.tsx
francisl: I put it in the middle so nobody can see it! 🤣 

Joke aside, I'll fix that for sure!

src/components/CohortBuilder/Summary/Cards/SummaryCard.tsx
francisl: in that case, the component should not know about it, since it's passing it around for the next component.

I know we have issue in the code about over using it. But I think It's a case where it make sense to do it.

PR #2694, opened 2020-09-11
Entrevue: feat: Split table in two for research and other member

francisl:
# DO NOT MERGE 
# Test

Acceptance Criteria

Separate normal user from researcher in two different list
Admin has log files, show the sum in the header of the member table
Admin table should have a custom border color of #23ab2f

### DIFF
--- a/src/components/MemberSearchPage/MemberSearchContainer.js
+++ b/src/components/MemberSearchPage/MemberSearchContainer.js
@@ -44,7 +44,7 @@ class MemberSearchContainer extends Component {
     members: PropTypes.array.isRequired,
   };
 
-  handleChange = e => {
+  handleChange = (e) => {
     const {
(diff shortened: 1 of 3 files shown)


src/components/MemberSearchPage/MemberSearchContainer.js
```diff
@@ -181,6 +181,36 @@ class MemberSearchContainer extends Component {
...
+
+    var othermembers = members
+      .map(function (m) {
+        if (m.roles.includes('research')) return null;
+        return m;
+      })
+      .filter(function (m) {
+        if (m !== null) return true;
+        return false;
+      });
```
francisl, line 207:
```javascript
var othermembers = members.filter((m) => {
    if (m.roles.includes('research')) return false;
        return true;
})
```

Issue #2615, opened 2020-07-08
Add Participants to an existing set


denis-crsj:
### Feature

Instead of saving the selection as a new Participants set, users should be able to add the selected participants to one of their existing saved set.

A new option called "Add to existing Participants set" need to be added to the dropdown menu under the "Save as new Participants set" option. If the user has at least one set already, the option will be clickable but if the user does not have any saved set, the option will be greyed out.

Selecting this option will open a Modal box with a dropdown field listing all his sets and allowing him to select the one to which he wants to add the participants.

It is extremely possible that a user might add a set of Participants to one that already contain some of the IDs in question. If it’s the case, no errors nor validations need to be displayed to the user, the system will run that validation in the background and only add the IDs that were not already in the chosen set.

### Acceptance Criteria

**Scenario 1** User has no saved set
  - **Given:** The user is on the Cohort Builder
    - **And:** he just ran a query displaying some results
  - **When:** the user clicks on the "Save Participants set" button
  - **Then:** the user should see 2 options in the dropdown. The "Save as new Participants set" option should be clickable and the "Add to existing Participants set" option should be greyed out and unclickable.

**Scenario 2** User has 1 or more saved sets (No duplicates)
  - **Given:** The user is on the Cohort Builder
    - **And:** he just ran a query displaying some results
  - **When:** the user clicks on the "Save Participants set" button
  - **Then:** The user should see 2 options [Save as new Participants set / Add to existing Participants set]
  - **When:** the user clicks on the "Add to existing Participants set" option, a modal box open displaying a dropdown menu containing the user's list of Saved sets.
  - **Then:** The user selects a set which _does not contain duplicates_ and clicks "Confirm/Save".
  - **Then:** The user goes back to the dashboard and the updated number of items in the set should be the sum of the old set and the new selection

**Scenario 3** User has 1 or more saved sets (With duplicates)
  - **Given:** The user is on the Cohort Builder
    - **And:** he just ran a query displaying some results
  - **When:** the user clicks on the "Save Participants set" button
  - **Then:** The user should see 2 options [Save as new Participants set / Add to existing Participants set]
  - **When:** the user clicks on the "Add to existing Participants set" option, a modal box open displaying a dropdown menu containing the user's list of Saved sets.
  - **Then:** The user selects a set which _does contain some duplicates_ and clicks "Confirm/Save".
[...]

PR #2701, opened 2020-09-14
fix(savesets): #2615 qa review

adipaul1981:
* quick fix(cohort builder, upload ids modal): fix layout

* fix lint


evans-g-crsj (approved):
I would remove comments since they can quickly become obsolete

PR #2723, opened 2020-09-22
fix: #2633 summary table css update

francisl:
#2633 

Screen Shot 2020-09-22 at 2 58 45 PM
Screen Shot 2020-09-22 at 2 58 49 PM


src/components/CohortBuilder/Results/Toolbar/Toolbar.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { roundIntToChosenPowerOfTen } from 'utils';
 import { Spin, notification } from 'antd';
-import familyMembers from 'assets/icon-families-grey.svg';
-import { FileFilled } from '@ant-design/icons';
+import DemographicIcon from 'icons/DemographicIcon';
+import FamilyMembersIcon from 'icons/FamilyMembersIcon';
+
+import { ArrowRightOutlined, FileFilled } from '@ant-design/icons';
 
 import ButtonWithRouter from '../../../../ui/ButtonWithRouter';
```
evans-g-crsj, line 12: we may not need the ../../

src/components/CohortBuilder/Summary/Cards/SummaryCard.css
```diff
@@ -27,6 +28,7 @@
   height: 100%;
   align-items: center;
   justify-content: center; */
```
evans-g-crsj, line 30: we should remove comment if not needed

src/icons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+};
+
+export const BaseSvg = ({
+  svg,
+  size = '1em',
+  height = size,
+  width = size,
+  style,
+  alt = '',
+  ...props
```
evans-g-crsj, line 30: I know that ...props is legitimate, here. But, I think we should not use and a needed prop when needed

evans-g-crsj (approved):

Issue #2750, opened 2020-09-29
Thin vertical bar to the right of the dashboard page


denis-crsj:
**Bug description**
The is a thin vertical bar to the right of the dashboard page (looking at the bottom scroll bar, there is a space)

**Steps To Reproduce**
1. Go to the Dashboard page
2. Scroll to the right
3. Notice there is a vertical white bar to the right of the page

**Expected behavior**
No vertical lines

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5ea980d6189dd52c91ddddfa/38f2acaf-9b35-4cfe-8c00-b1571340ea96)

**Additional context**

PR #2752, opened 2020-09-29
fix(dashboard): #2750 remove horizontal scroll in dashboard

adipaul1981:


src/components/UserDashboard/UserDashboard.module.css
```diff
@@ -1,3 +1,8 @@
+.dashboadRowContainer {
+  margin: 0 !important;
```
francisl, line 2: I know `!important` is valid css. But it's use should be the last-last resort when there is no other possibilities.

What did you tried to force the use of it? Can't you make your selector more specific?

francisl (changes requested):

evans-g-crsj (approved):

PR #2802, opened 2020-10-22
2800 nav button

evans-g-crsj:


adipaul1981 (approved):

src/components/Header/index.js
```diff
@@ -1,6 +1,5 @@
-import React, { Fragment } from 'react';
+import React from 'react';
```
adipaul1981, line 1: The only comment i have is that the button "move" on mouse over. This should be fixed as it's an irritant. 

image

Issue #2642, opened 2020-08-07
Add a link to the Sunburst card to add the selected HPO term to the active query


denis-crsj:
### Feature

We want to add a link that would allow the user to add the term that is currently selected in the Sunburst to the active query in the Cohort builder.

![image](https://user-images.githubusercontent.com/64538118/97216224-e38d7380-179b-11eb-8c37-e10183434f54.png)

### Acceptance Criteria

N/A

### Sub-Tasks
_List of what needs to be done  on the technical side._
- [ ] Sub-task 1

### Testing Strategy
_What do we need to test? Who can test it? How do we test it? What do we need to test it?_

### Defects
_List of defects/bugs found during QA._
- [ ] Defect 1

### Resources

Sunburst specs: https://www.notion.so/ferlab/KF-Ontology-Sunburst-212f8110b7794cbca44198a1ab7f993b

Figma designs: https://www.figma.com/file/0voQuoBD3Xvap4wYQ0VLYq/KF---Sunburst?node-id=151%3A0

denis-crsj:
Tested successfully. We will have to discuss and implement the correct behavior though when adding the term in the query from the sunburst card.

PR #2815, opened 2020-11-02
feat(sunburst): #2642 add link from sunburst to cohort builder

adipaul1981:


src/common/sqonUtils.js
```diff
@@ -221,3 +232,35 @@ export const addSetToActiveQuery = ({ setId, querySqons, activeIndex }) => {
   newSqons[activeIndex] = activeSqon;
   return newSqons;
 };
+
+export const addFieldToActiveQuery = ({ term, querySqons, activeIndex }) => {
+  if (!term) {
```
evans-g-crsj, line 237: I would remove that check si it's the caller responsibility to pass correct term.

src/common/sqonUtils.js
```diff
@@ -221,3 +232,35 @@ export const addSetToActiveQuery = ({ setId, querySqons, activeIndex }) => {
...
+  const sameTermIndex = activeSqon.findIndex((s) => s.content.field === term.field);
+  if (sameTermIndex >= 0) {
+    let valuesAsSet = new Set();
+    activeSqon[sameTermIndex].content.value.forEach((v) => valuesAsSet.add(v));
+    valuesAsSet.add(term.value);
+    return activeSqon.map((s) => {
+      if (s.content.field === term.field) {
+        return { ...s, content: { field: term.field, value: [...valuesAsSet] } };
+      } else {
+        return s;
```
evans-g-crsj, line 260: should return a copy of s, I think, {...s}

src/common/sqonUtils.js
```diff
@@ -221,3 +232,35 @@ export const addSetToActiveQuery = ({ setId, querySqons, activeIndex }) => {
   newSqons[activeIndex] = activeSqon;
   return newSqons;
 };
+
+export const addFieldToActiveQuery = ({ term, querySqons, activeIndex }) => {
+  if (!term) {
+    return querySqons;
+  }
+
+  const activeSqon = [...querySqons[activeIndex].content];
```
evans-g-crsj, line 241: no need to copy, since you only query the variable in addFieldToActiveSqon

src/common/sqonUtils.js
```diff
@@ -221,3 +232,35 @@ export const addSetToActiveQuery = ({ setId, querySqons, activeIndex }) => {
...
+  const newQuerySqon = [...querySqons];
+
+  newQuerySqon[activeIndex] = { op: newQuerySqon[activeIndex].op, content: newActiveSqon };
+  return newQuerySqon;
+};
+
+const addFieldToActiveSqon = (term, activeSqon) => {
+  const sameTermIndex = activeSqon.findIndex((s) => s.content.field === term.field);
+  if (sameTermIndex >= 0) {
+    let valuesAsSet = new Set();
```
evans-g-crsj, line 253: I think you could use const vs let

src/common/sqonUtils.js
evans-g-crsj, line 241: naming: it's not a sqon it's sqon content

evans-g-crsj (changes requested):

evans-g-crsj (approved):

PR #2820, opened 2020-11-06
Feat/2641 hpo tree in sb

adipaul1981:
image


src/components/OntologyBrowser/Test/store.test.ts
```diff
@@ -1,6 +1,13 @@
...
+import {
+  generateNavTreeFormKey,
+  PhenotypeSource,
+  PhenotypeStore,
+  removeSameTerms,
+  selectSameTerms,
+} from '../store';
 import { TreeNode } from '../Model';
 import { flatMockData, treeData } from './mockData';
+import { hpoTreeTitleFormat } from '../../UI/Charts/Sunburst/Sunburst';
```
francisl, line 10: `import { hpoTreeTitleFormat } from 'components/UI/Charts/Sunburst/Sunburst';`

src/components/OntologyBrowser/store.ts
```diff
@@ -1,5 +1,7 @@
 import { graphql } from '../../services/arranger';
 import OntologyTree, { TreeNode } from './Model';
+import { hpoTreeTitleFormat } from '../UI/Charts/Sunburst/Sunburst';
```
francisl, line 3: absolute import ☝️

src/components/OntologyBrowser/store.ts
```diff
@@ -10,6 +12,15 @@ export type PhenotypeSource = {
   filter_by_term: any;
 };
 
+export type TreeNodeLight = {
+  key: string;
+  title: React.ReactElement | string;
+  children: TreeNode[];
+};
+
```
francisl, line 20: I'm not convinced that a new type is the way to go here. That will make the Node incompatible with other process and there is not that much difference.

I think it would be better to have a type constructor that return a built 'light' object. (set the default)

And please put the type definition in the model file.

src/components/OntologyBrowser/store.ts
```diff
@@ -61,6 +72,47 @@ const findAllSameTerms = (
...
+  }
+  const rootPheno = phenotypes.pop();
+  const splitPheno = splitHPOTerm(rootPheno || '');
+
+  return [
+    {
+      key: rootPheno,
+      title: hpoTreeTitleFormat(splitPheno, selectedPhenotypeInfoTitle),
+      children: generateNavTreeFormKey(phenotypes, selectedPhenotypeInfoTitle),
+    } as TreeNode,
```
francisl, line 103: line 84-85 and 95-96 are exactly the same. If you use a type/object constructor, this would reduce the logic here.

src/components/OntologyBrowser/store.ts
```diff
@@ -61,6 +72,47 @@ const findAllSameTerms = (
...
+  ];
+};
+
+export const splitHPOTerm = (term: string) => {
+  let match;
+  match = RegexHPOCode.exec(term);
```
francisl, line 109: one liner?

src/components/UI/Charts/Sunburst/InfoPanelNavTree.tsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable react/prop-types */
+import React, { FunctionComponent } from 'react';
+import './sunburst.css';
```
francisl, line 3: I know it's not in our guideline, but what do you think if we put all css import at the end of 'js' imports?

src/components/UI/Charts/Sunburst/Sunburst.tsx
```diff
@@ -22,10 +34,26 @@ type SunburstProps = {
...
 const pickPhenotypeInfo = (p: Phenotype) => ({
   title: p.title,
+  key: p.key,
   results: p.results,
   exactTagCount: p.exactTagCount,
 });
 
+export const hpoTreeTitleFormat = (splitPheno: PhenotypeSplit, currentTerm: string) => {
+  const currentPhenoCode = splitHPOTerm(currentTerm).code;
+  return (
```
francisl, line 44: it's only use one time, maybe a no return function will work here. (your call)

src/components/UI/Charts/Sunburst/Sunburst.tsx
```diff
@@ -22,10 +34,26 @@ type SunburstProps = {
...
+    <div
+      className={'hpo-tree-name'}
+      style={{
+        fontWeight: currentPhenoCode !== splitPheno.code ? 400 : 'bold',
```
francisl, line 48: I think we should avoid inline, use class instead

francisl (changes requested):

src/components/OntologyBrowser/store.ts
```diff
@@ -61,6 +72,47 @@ const findAllSameTerms = (
...
+  return [
+    {
+      key: rootPheno,
+      title: hpoTreeTitleFormat(splitPheno, selectedPhenotypeInfoTitle),
+      children: generateNavTreeFormKey(phenotypes, selectedPhenotypeInfoTitle),
+    } as TreeNode,
+  ];
+};
+
+export const splitHPOTerm = (term: string) => {
```
evans-g-crsj, line 107: "const match" should work + hoist function before its use

src/components/OntologyBrowser/store.ts
```diff
@@ -61,6 +72,47 @@ const findAllSameTerms = (
   return sameTermKeys;
 };
 
+export const generateNavTreeFormKey = (
+  phenotypes: string[],
+  selectedPhenotypeInfoTitle: string,
+): TreeNodeLight[] => {
+  if (!phenotypes.length) {
```
evans-g-crsj, line 79: is that check needed?

src/components/OntologyBrowser/store.ts
```diff
@@ -61,6 +72,47 @@ const findAllSameTerms = (
...
+    return [];
+  }
+
+  if (phenotypes.length === 1) {
```
evans-g-crsj, line 83: I think we should avoid mutating when possible. For example myArray.slice(0, -1) vs myArray.pop()

src/components/UI/Charts/Sunburst/InfoPanel.tsx
```diff
@@ -62,8 +64,10 @@ const InfoPanel: FunctionComponent<Props> = ({ data, onClickAddTermToActiveIndex
         </div>
       </div>
       <div className={'tree-grid'}>
-        <div className={'tree-title'}>Current Path</div>
-        <div className={'tree'}>tree to come</div>
+        <div className={'tree-title'} style={{ paddingTop: 12 }}>
```
evans-g-crsj, line 67: There is a css file already  so css vs style if possible

src/components/UI/Charts/Sunburst/InfoPanelNavTree.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Tree } from 'antd';
+import { TreeNode } from '../../../OntologyBrowser/Model';
+
+type OwnProps = {
+  treeData: TreeNode[];
+  expandedKeys: string[];
+};
+
+const InfoPanelNavTree: FunctionComponent<OwnProps> = ({ treeData, expandedKeys }) => (
```
evans-g-crsj, line 12: Why do we need this component? it seems like a very thin layer above antd tree?

src/components/UI/Charts/Sunburst/Sunburst.tsx
```diff
@@ -22,10 +34,26 @@ type SunburstProps = {
...
+      }}
+    >
+      {splitPheno.name}
+      <span className={'hpo-tree-code'}>{' ' + splitPheno.code}</span>
```
evans-g-crsj, line 52: template string

src/components/UI/Charts/Sunburst/Sunburst.tsx
evans-g-crsj, line 48: I would use css with a condition vs style and put my condition in a var => 
const isCurrentPhenoCode = currentPhenoCode !== splitPheno.code;

className={'hpo-tree-name ${isCurrentPhenoCode ? 'myClassName': ''}'}

src/components/UI/Charts/Sunburst/Sunburst.tsx
```diff
@@ -41,12 +69,18 @@ class Sunburst extends Component<SunburstProps, State> {
 
   state = {
     selectedPhenotypeInfo: pickPhenotypeInfo(this.props.data),
+    phenotypeTree: [],
   };
 
-  getSelectedPhenotype = (phenotype: Phenotype) =>
+  getSelectedPhenotype = (phenotype: Phenotype) => {
+    const phenotypes = phenotype.key.match(RegexExtractPhenotype);
+    const phenoReversed = (phenotypes || []).reverse();
```
evans-g-crsj, line 77: Note: reverse() modifies the array. It can be ok. Just wanna let you know in case it's not.

evans-g-crsj (approved):

francisl (approved):
:shipit:

francisl:
just make sure all the tests are green

PR #2847, opened 2020-11-30
Feat/clickable tree

adipaul1981:
image

### DIFF
--- a/src/components/UI/Charts/Sunburst/InfoPanel.tsx
+++ b/src/components/UI/Charts/Sunburst/InfoPanel.tsx
@@ -9,7 +9,7 @@ import { AddTermToActiveIndex, Term } from 'store/virtualStudiesTypes';
 import { Button, Tree } from 'antd';
 import { TreeNode } from 'components/OntologyBrowser/Model';
 import './sunburst.css';
-import { RegexExtractPhenotype } from '../../../OntologyBrowser/store';
+import { RegexExtractPhenotype } from 'components/OntologyBrowser/store';
 
 type OwnProps = {
   data: Pick<Phenotype, 'title' | 'key' | 'results' | 'exactTagCount'>;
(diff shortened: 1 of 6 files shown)


src/components/UI/Charts/Sunburst/InfoPanel.tsx
```diff
@@ -9,10 +9,12 @@ import { AddTermToActiveIndex, Term } from 'store/virtualStudiesTypes';
 import { Button, Tree } from 'antd';
 import { TreeNode } from 'components/OntologyBrowser/Model';
 import './sunburst.css';
+import { RegexExtractPhenotype } from '../../../OntologyBrowser/store';
```
evans-g-crsj, line 12: is ../../.. needed?

src/components/UI/Charts/Sunburst/InfoPanel.tsx
```diff
@@ -29,11 +31,36 @@ const splitTitle = (title: string) => {
...
     title: rawTitle.trim(),
-    hpCode: `HP ${rawCode.replace(')', '').trim()}`,
+    hpCode: rawCode ? `HP ${rawCode.replace(')', '').trim()}` : '',
   };
 };
 
-const InfoPanel: FunctionComponent<Props> = ({ data, onClickAddTermToActiveIndex, treeData }) => {
+export const getPath = (node: string, treeNodes: TreeNode[], path: string[] = []) => {
+  const currentNodeText = treeNodes[0].key;
+  path.push(currentNodeText);
```
evans-g-crsj, line 40: here, the param is mutated. If possible, I would make a shallow copy of "path" (concat, slice, ...)

src/components/UI/Charts/Sunburst/Sunburst.tsx
```diff
@@ -120,7 +128,11 @@ class Sunburst extends Component<SunburstProps, State> {
           ref={this.ref}
         />
         <div className={'grid-item'}>
-          <InfoPanel data={{ ...selectedPhenotypeInfo }} treeData={phenotypeTree} />
+          <InfoPanel
+            data={{ ...selectedPhenotypeInfo }}
```
evans-g-crsj, line 132: I don't think that copying here is needed since props are copied (to be verified)

src/components/UI/Charts/Sunburst/sunburst-d3.jsx
```diff
@@ -265,6 +265,25 @@ const sunburstD3 = (ref, data, config, getSelectedPhenotype, formatters) => {
     const textData = p || selectedPhenotype;
     centerText.text(() => centerTextFormatter(textData.data)).call(wrap);
   };
+
+  const findNodeByKey = (key, node, returnNode) => {
```
evans-g-crsj, line 269: can this function be extracted on it's own? 

instead of looping through all nodes, could "some" or "find" do the job?

evans-g-crsj (approved):

src/components/UI/Charts/Sunburst/InfoPanel.tsx
adipaul1981: in this case i think it ok that its mutated. The initial value is [] anyway.

Issue #2851, opened 2020-12-01
Cohort Builder: Suggest better default titles for saved sets


vferretti:
Currently, default titles are in format Saved_set_n. Implement same default title rules as for the File Repo saved queries.

Don't provide any default title for combined queries, i.e. queries that refers to another one (e.g #Q1 U #Q2)

denis-crsj:
De mon côté tout semble beau.

adipaul1981:
ca me semble ok aussi.

PR #2859, opened 2020-12-03
feat(sets): #2851 set default name as per file repo

adipaul1981:
image


src/common/Test/sqonToName.test.js
```diff
@@ -0,0 +1,24 @@
+import { expect } from 'chai';
+import { cyan } from 'chalk';
```
evans-g-crsj, line 2: Nice to know!

evans-g-crsj (approved):

PR #2906, opened 2021-01-18
Feat/2437 ontologybrowser add any of/all of/notin modal

adipaul1981:
image


src/components/OntologyBrowser/index.tsx
```diff
@@ -44,7 +49,44 @@ const desactivateAllSameTerms = (allSameTerms: string[], transferItems: Transfer
     } else return t;
   });
 
-export const updateSqons = (initialSqon: Sqon, value: string[], selectedField: string): Sqon => {
+const opComponent = (fill: string, operation: string) => {
```
evans-g-crsj, line 52: maybe change the var name that includes the fact that it returns an icon: eg, "displayOpIcon" or something similar.

src/components/OntologyBrowser/index.tsx
```diff
@@ -44,7 +49,44 @@ const desactivateAllSameTerms = (allSameTerms: string[], transferItems: Transfer
...
+    operation: ALL_OP,
+  },
+  {
+    name: 'Not',
+    component: (fill: string) => opComponent(fill, SOME_NOT_IN_OP),
+    operation: SOME_NOT_IN_OP,
+  },
+];
+
+const getOperations = (name: string) =>
```
evans-g-crsj, line 81: I think we should remove the 's' since it returns 1 operation

src/components/OntologyBrowser/index.tsx
```diff
@@ -44,7 +49,44 @@ const desactivateAllSameTerms = (allSameTerms: string[], transferItems: Transfer
...
+      return <NotEqualIcon fill={fill} />;
+    default:
+      return <UnionIcon fill={fill} />;
+  }
+};
+
+const OPERATIONS = [
+  {
+    name: 'Any of',
+    component: (fill: string) => opComponent(fill, IN_OP),
```
evans-g-crsj, line 66: do we really need the opComponent func? Could we just return the desired icon of a given name?

src/components/OntologyBrowser/index.tsx
```diff
@@ -211,14 +254,29 @@ class OntologyModal extends React.Component<ModalProps, ModalState> {
...
   isLoadingOrEmpty = (): boolean => this.state.isLoading || this.transfertDataSource.length === 0;
 
+  menu = () => (
+    <Menu
+      onClick={(e) => {
+        this.setState({ selectedOperator: e.key as string });
+      }}
+    >
+      {OPERATIONS.map((s) => (
+        <Menu.Item key={s.name.toLowerCase()} icon={s.component('#515885')}>
```
evans-g-crsj, line 264: is the .toLowerCase() necessary?

evans-g-crsj (approved):

src/components/OntologyBrowser/index.tsx
adipaul1981: this is just to satisfy esLint...

src/components/OntologyBrowser/index.tsx
adipaul1981: Well one could make sure that both sides are identical lower/upper cased. But i think it avoids any potential errors, since we are really just interested of the text content (case insensitive)

evans-g-crsj (approved):

Issue #2914, opened 2021-01-21
Update numbers in variant page


jecos:
Number are out of date. Here the updated numbers : 
```
studies=6
participantsCount = 5053
distinctVariantsCount = 200569156
occurrencesCount = 21295117739
```

PR #2915, opened 2021-01-21
fix: update numbers for varaint db (#2914)

jecos:


src/components/VariantDb/index.tsx
```diff
@@ -229,7 +229,7 @@ class VariantDb extends React.Component<VariantDbProps, VariantDbState> {
                         color: '#7D84A6',
                       }}
                     >
-                      May 13, 2020
+                      May 21  , 2021
```
botekchristophe, line 232: there few extra spaces before the coma
shouldn't be 'January 21, 2021' instead ?

botekchristophe (approved):

PR #2948, opened 2021-02-16
fix: #2947 new @ferlab/ui

francisl:


package.json
```diff
@@ -3,8 +3,8 @@
   "version": "2.34.0",
   "private": true,
   "dependencies": {
-    "@ferlab-ui/core-react": "0.0.7",
-    "@ferlab-ui/core-style": "0.0.8",
+    "@ferlab/ui": "0.1.0",
```
evans-g-crsj, line 6: I know there were a visual bug (when cards are in loading state, the loader is shrunk). Was it corrected?

evans-g-crsj (approved):

package.json
francisl: it's a name refactor and a builder process fix. Not feature/bugfix PR

PR #2966, opened 2021-02-20
2947 fix undefined routes

francisl:


src/App.js
```diff
@@ -193,6 +193,37 @@ const App = compose(
...
+          path={ROUTES.studies}
+          exact
+          render={(props) =>
+            protectRoute({
+              api,
+              isLoadingUser,
+              Component: () => {
+                import('pages/studies')
+                  .then(({ StudiesPage }) => StudiesPage)
+                  .catch(() => {});
```
evans-g-crsj, line 220: is the "catch" necessary?

src/App.js
```diff
@@ -193,6 +193,37 @@ const App = compose(
...
+              },
+              loggedInUser,
+              ...props,
```
evans-g-crsj, line 223: if possible, let's avoid the ...props

evans-g-crsj (approved):

src/App.js
francisl: I'll add handling the catch clause in next update

src/App.js
francisl: in next update

src/App.js
```diff
@@ -193,6 +193,37 @@ const App = compose(
...
+              api,
+              isLoadingUser,
+              Component: UserDashboard,
+              loggedInUser,
+              ...props,
+            })
+          }
+        />
+
+        <Route
```
evans-g-crsj, line 210: shouldn't that be wrapped in a feature toggle too?

evans-g-crsj (changes requested):

evans-g-crsj (approved):

PR #2981, opened 2021-03-04
feat: #2959 Apollo integration with studies

francisl:


src/store/providers/tableColumns.ts
```diff
@@ -0,0 +1,28 @@
+import { makeVar } from '@apollo/client';
+
+const getTableColumns = (): Record<string, any> =>
+  JSON.parse(localStorage.getItem('tableColumns') || '{}') || {};
```
evans-g-crsj, line 4: is the last || {} needed?

src/store/providers/tableColumns.ts
```diff
@@ -0,0 +1,28 @@
...
+export const configureColumns = (original: Array<any>, newOrder: Array<any>): Array<any> => {
+  const newDataOrder: Array<any> = [];
+  newOrder.forEach((newValue) => {
```
evans-g-crsj, line 7: I think that this forEach loop is absolutely correct. But personally, I would favor a more functional approach (reduce for instance)

evans-g-crsj (approved):

src/components/Layout/PageContent.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { FC } from 'react';
+import PageTitle from 'components/Text/PageTitle';
+
+import style from './PageContent.module.scss';
+
+type PageContentProps = {
+  title: string | void;
```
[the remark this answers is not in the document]
francisl: care to explain. This is mean to be a reusable component instead of replicating the same css/html in our different page with different title

evans-g-crsj (approved):

PR #2982, opened 2021-03-04
feat (variant page): minimal UI

evans-g-crsj:
image


src/pages/variants/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+import StackLayout from '@ferlab/ui/core/layout/StackLayout';
+import { Typography } from 'antd';
+import style from './Variants.module.scss';
+import SearchView from './SearchView';
+
+const { Title } = Typography;
+
+const VariantPage: FunctionComponent = () => {
+  return (
+    <StackLayout vertical className={style.mainContainer}>
```
francisl, line 11: maybe PageContent will reduce the css needed here (and remove the stack layout

src/pages/variants/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <div className={style.titleContainer}>
```
francisl, line 12: These title are use on all page, they shouldn't have to require so much css on each page. I think the PageContent will also fix that

src/pages/variants/VariantTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { Table } from 'antd';
+import style from './VariantTable.module.scss';
+import { Spinner } from '../../uikit/Spinner';
+
+type OwnProps = {
+  isLoading: boolean;
+  dataSource: any[];
+};
+
+const columns = [
```
francisl, line 11: if index and key are always the same, I would use a loop to add them dynamically, one typo and you'll have a bug hard to fix

src/pages/variants/SearchOption.module.scss
```diff
@@ -0,0 +1,51 @@
...
+  border-radius: 2px;
+  font-weight: normal;
+  font-size: 12px;
+  line-height: 20px;
+  letter-spacing: 0.25px;
+  color: #2b388f; //TODO var
+}
+
+.leftSquareBgColorVariant {
+  background: #e6bddc;
```
francisl, line 27: isn't that part of the theme or anti default?

src/pages/variants/SearchOption.tsx
```diff
@@ -0,0 +1,56 @@
...
+    return style.leftSquareBgColorGene;
+  }
+  return style.leftSquareBgColorVariant;
+};
+
+const SearchOption: FunctionComponent<OwnProps> = (props) => {
+  const { criteria, searchParams } = props;
+  const { symbols, locus } = searchParams;
+  return (
+    <div className={style.container}>
```
francisl, line 38: two layer of StackLayout here, it will remove the need to add custom css for the flex stuff.
If you find issues with the stackLayout, please provide feedback or update to make them better.

francisl (changes requested):

src/pages/variants/SearchView.module.scss
```diff
@@ -0,0 +1,20 @@
+@import 'src/style/themes/default/colors.module.scss';
```
francisl, line 1: please look at other import for Colors

src/pages/variants/SuggestionOptions.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const generateDisplayName = (suggestion: Suggestion): string => {
+  const type = suggestion.type;
+  if (type === GenomicFeatureType.Variant) {
+    return suggestion.locus || '';
+  } else if (type === GenomicFeatureType.GENE) {
+    return suggestion.suggestion_id; // gene symbol.
+  }
+  return '';
+};
```
francisl, line 14: you can also do like you already did in this pr. it looks clean.

```const suggestions = {
  [GenomicFeatureType.Variant]: suggestion.locus,
  [GenomicFeatureType.GENE]: suggestion.suggestion_id 
}
return suggestions[suggestion.type] || ''
```

src/store/graphql/variants/actions.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const findQuery = (type: GenomicFeatureType) => {
+  if (type === GenomicFeatureType.GENE) {
+    return GENE_TABLE_QUERY;
+  } else if (type === GenomicFeatureType.Variant) {
+    return VARIANT_TABLE_QUERY;
+  }
+  return VARIANT_TABLE_QUERY;
+};
```
francisl, line 18: why `else if` if it's the default?

francisl (approved):
LGTM

src/pages/variants/ConsequencesCell.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { FunctionComponent } from 'react';
```
francisl, line 1: We should reconsider using that pattern.

see . https://github.com/facebook/create-react-app/pull/8177

This is not a blocker since we use it elsewhere. But we should avoid it in next PR.

Please for the moment use { FC } instead. It will easy the future refactoring since it's how we use it normally.

src/pages/variants/Stats.tsx
```diff
@@ -0,0 +1,5 @@
+import React, { FunctionComponent } from 'react';
```
francisl, line 1: please use FC. It's the standard we use in KF and other project.

src/pages/variants/SuggestionOption.module.scss
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.detailsContainer {
+  margin: 0 12px;
+}
+
+.displayName {
+  font-weight: 600;
+  font-size: 14px;
+  line-height: 22px;
```
francisl, line 34: This is standard in the project, we should not need to specify it.

src/pages/variants/VariantTable.module.scss
```diff
@@ -0,0 +1,12 @@
+.table {
+  width: 100%;
+}
+
+.link {
+  text-decoration-line: underline;
+  color: #383f72;
```
francisl, line 7: why are link different on this page?

src/store/genomicSuggesterTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+interface ReInitializeState {
+  type: typeof RE_INITIALIZE_STATE;
+}
+
+interface ClearSuggestions {
+  type: typeof CLEAR_SUGGESTIONS;
+}
+
+interface ToggleLoading {
+  type: typeof TOGGLE_LOADING;
```
francisl, line 37: aren't all your type 'string' if you use `typeof` ? isn't a enum will be clearer in that case? Why is it a interface and not a type? I can't see any object implanting the interface.

src/store/graphql/variants/models.ts
```diff
@@ -0,0 +1,62 @@
...
+    canonical?: boolean;
+    aa_change: string | undefined;
+    [key: string]: any;
+  };
+  [key: string]: any;
+};
+
+export type ClinVar = {
+  clinvar_id: string | undefined;
+  clin_sig: string | undefined;
```
francisl, line 22: `?.` and `T | undefined` is exactly the same definition. We should be consistant accros the project yet the same file.

src/style/themes/default/antd/antd.less
```diff
@@ -47,3 +47,20 @@ h1.ant-typography {
...
+.ant-table-cell {
+  font-weight: normal;
+  font-size: 14px;
+  /* Body / body-1 TODO */
+  color: #383F72;
+}
+.ant-table-cell > a {
+  text-decoration-line: underline;
+  color: #383F72; // TODO var
+}
```
francisl, line 66: There is a `table.less` file for ant related

francisl (changes requested):
There is many easy fix that will be forgotten if we don't fix now

src/services/genomicSuggestions.js
```diff
@@ -0,0 +1,7 @@
+import { apiInitialized } from './api';
+
+export const getGenomicSuggestions = async (searchText) =>
+  await apiInitialized({
+    url: `https://kf-api-arranger-qa.kf-strides.org/genomicFeature/suggestions/${searchText}`, //TODO
```
adipaul1981, line 5: I would resole the TODO before pushing into QA here. Its easy to forget.

adipaul1981 (changes requested):

src/store/genomicSuggesterTypes.ts
```diff
@@ -0,0 +1,57 @@
...
+  type: typeof FAILURE;
+  error: Error | null;
+};
+
+type ReInitializeState = {
+  type: typeof RE_INITIALIZE_STATE;
+};
+
+type ClearSuggestions = {
+  type: typeof CLEAR_SUGGESTIONS;
```
francisl, line 33: I don't understand that.

why do you create a bunch of type with a type of type `'string'`. Why type is not an enum? that way, each type will be unique.

adipaul1981 (approved):

PR #2994, opened 2021-03-23
fix: #2993 new project

adipaul1981:


src/store/graphql/studies/queries.tsx
```diff
@@ -3,7 +3,7 @@ import { fields } from './models';
 
 export const STUDIES_QUERY = gql`
   query StudiesInformation ($sqon: JSON){
-    study {
+    Studies {
```
evans-g-crsj, line 6: Maybe prepare the query for paging. Since we know that shortly we will use pagination.

evans-g-crsj (approved):

francisl (approved):
LGTM :shipit:

Issue #2999, opened 2021-03-24
Update default button theme (antd)


francisl:
https://www.figma.com/file/cnK68RMYFGoLbgsnCYvoo6/FerlabUI?node-id=647%3A10344

PR #3003, opened 2021-03-26
fix: #2999 Fix Querybuilder and button default theme

francisl:


src/style/themes/default/antd/buttons.less
```diff
@@ -0,0 +1,85 @@
...
+    &:focus {
+      > span {
+        text-decoration: none;
+      }
+    }
+  }
+
+  // &:focus {
+  //   & when (@theme = dark) {
+  //     .button-color(
```
adipaul1981, line 65: remove?

adipaul1981 (approved):

PR #3021, opened 2021-03-31
WIP: connect to cluster component

evans-g-crsj:
https://portal-qa.kidsfirstdrc.org/devVariantDb

image


src/services/workBench.js
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+export const stopCluster = async () =>
+  await api({
+    method: 'DELETE',
+    url: kfVariantClusterUrl,
+  });
+
+export const getClusterStatus = async () =>
+  await api({
```
francisl, line 19: Why is it required to block the thread here if it's pass to a thunk?

francisl (commented):
Looking good.
I do have a question though about the await.

src/services/workBench.js
evans-g-crsj: fixed!

Issue #3004, opened 2021-03-26
Cohort Builder - Participant table: Replace column "Study Name" by "Study Code"


vferretti:
The column title could be simply "Study". Do the same in the File Repo

Issue #3001, opened 2021-03-24
Add facets to the CB and FR 


vferretti:
We need to add the following facets in the Cohort Builder:

- Program ( in menu Study), 
- Domain (in menu  Study)
- Data Category (in menu  Available Data)

So that the menu study will be ((in that order)
Study
Program
Domain
Data Access Authority
Proband
 
and the menu Available Data would be
Available Data Category
Available Data Types
Experiment Strategy
Family Shared Available Data Types

Add same facets to the File Repo
Program et Domain dans l'onglet Clinical Filter en dessous de Study name
Data Category dans l'onglet File FIlters en dessous de Harmonized data

PR #3022, opened 2021-04-01
Add filters to cb and file repo

adipaul1981:
Resolves both [https://github.com/kids-first/kf-portal-ui/issues/3004](url) & [https://github.com/kids-first/kf-portal-ui/issues/3001](url)

image

image

image

image

image

### DIFF
--- a/src/components/FileRepo/AggregationSidebar/aggsConfig.js
+++ b/src/components/FileRepo/AggregationSidebar/aggsConfig.js
@@ -5,6 +5,8 @@
 // Fields that will be displayed in the "Clinical Filters" section
 export const CLINICAL_FILTERS = [
   'participants__study__short_name',
+  // 'participants__study__domain',
+  // 'participants__study__program',
   'participants__diagnoses__diagnosis_category',
   'participants__diagnoses__source_text_diagnosis',
   'participants__phenotype__hpo_phenotype_observed_text',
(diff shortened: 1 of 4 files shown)


evans-g-crsj (approved):

src/components/FileRepo/AggregationSidebar/aggsConfig.js
```diff
@@ -33,11 +36,11 @@ export const aggsConfig = (data, graphqlField) => {
...
     .fields;
 
-  const typeAggsConfig = type =>
-    type.map(fieldName => ({
+  const typeAggsConfig = (type) =>
+    type.map((fieldName) => ({
       field: fieldName,
       show: true,
-      type: gqlAggregationFields.find(fileAggField => fieldName === fileAggField.name).type.name,
+      type: gqlAggregationFields.find((fileAggField) => fieldName === fileAggField.name).type.name,
```
francisl, line 43: I don't know if you want to take the opportunity to rename 'type'. 
Just the fact that it's singular, so how can a `.map` work on it.

francisl (approved):
I leave a comment I would suggest you take the opportunity to make the semantic clearer. But I won't block the PR over this

src/components/FileRepo/AggregationSidebar/aggsConfig.js
evans-g-crsj: Good point. I would add that .find(fileAggField => fieldName === fileAggField.name).type.name, could be extracted in 1 func

PR #3090, opened 2021-04-13
Feat/3084 variant search page link to cb for studies

adipaul1981:
case 1:
image

case 2:
image

### DIFF
--- a/src/pages/variantsSearchPage/VariantStats.tsx
+++ b/src/pages/variantsSearchPage/VariantStats.tsx
@@ -26,7 +26,7 @@ const VariantStatsContainer = () => {
       title={
         <div className={style.variantStatsContainer}>
           <Title level={3}>Data Release 1</Title>
-          <div>April 08th, 2021</div>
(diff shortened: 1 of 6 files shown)


src/pages/variantsSearchPage/VariantStats.tsx
```diff
@@ -26,7 +26,7 @@ const VariantStatsContainer = () => {
       title={
         <div className={style.variantStatsContainer}>
           <Title level={3}>Data Release 1</Title>
-          <div>April 08th, 2021</div>
+          <div>January 21th, 2021</div>
```
francisl, line 29: why is it hardcoded?

evans-g-crsj (approved):

francisl (approved):

src/pages/variantsSearchPage/VariantStats.tsx
adipaul1981: we dont have the info in the database. we have to add a timestamp

Issue #3110, opened 2021-04-14
Remove participant link in variant search page and in variant entity page


adipaul1981:
Remove all links from search variant page and Variant entity page.

PR #3115, opened 2021-04-14
fix: #3110 remove participants links

adipaul1981:


evans-g-crsj (changes requested):
we should not leave commented out code.

evans-g-crsj (approved):
we should not leave commented out code.

PR #3153, opened 2021-04-26
Feat/2905: Display a specific message in case of no open connection to data integrations

celinepelletier:
This Pull request is a Draft because i wait for the final mockup and texts.

It also requires : https://github.com/kids-first/kf-api-variant-cluster/pull/36 to be merged

Capture d’écran, le 2021-04-26 à 11 19 02


src/pages/variantsSearchPage/WorkBench.tsx
```diff
@@ -125,29 +134,61 @@ const WorkBench = (props: Props) => {
...
-            )}
+            {error && displayError(error, loggedInUser, onReInitializeState)}
           </div>
         </StackLayout>
       </StackLayout>
     </Card>
   );
 };
 
+const displayError = (error: any, loggedInUser: LoggedInUser, onReInitializeState: () => void) => {
```
francisl, line 145: the type of error is defined in `store/WorkBenchState.ts` We should avoid any except when we have good reason to and document it in the code.

francisl (changes requested):

francisl (approved):
:shipit:

Issue #3134, opened 2021-04-20
Match Querybar and filter names


francisl:
Make sure that facet names are the same as the ones displayed in the query bar. 
Currently, the facet Domain is in lower case in the query i.e. domain = Cancer

PR #3155, opened 2021-04-26
fix: #3134 Match Querybar and filter names

celinepelletier:
Make sure that facet names are the same as the ones displayed in the query bar.
Currently, the facet Domain is in lower case in the query i.e. domain = Cancer

Capture d’écran, le 2021-04-26 à 16 07 33

### DIFF
--- a/src/pages/studies/StudyPageContainer.tsx
+++ b/src/pages/studies/StudyPageContainer.tsx
@@ -17,20 +17,19 @@ import { IDictionary } from '@ferlab/ui/core/components/QueryBuilder/types';
 
 type StudyPageContainerProps = StudiesPageContainerData & PaginationType;
 
-const StudyPageContainer = ({ studiesResults, filters, pagination }: StudyPageContainerProps) => {
+const StudyPageContainer = ({
+  studiesResults,
+  studiesMappingResults,
+  filters,
+  pagination,
+}: StudyPageContainerProps) => {
   const total = studiesResults?.data?.hits?.total || 0;
-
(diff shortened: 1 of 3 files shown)


src/pages/studies/StudyPageContainer.tsx
```diff
@@ -13,12 +13,27 @@ import { StudiesPageContainerData } from '../../store/graphql/studies/actions';
...
 
 import styles from './StudiesPageContainer.module.scss';
+import { IDictionary } from '@ferlab/ui/core/components/QueryBuilder/types';
 
 type StudyPageContainerProps = StudiesPageContainerData & PaginationType;
 
 const StudyPageContainer = ({ studiesResults, filters, pagination }: StudyPageContainerProps) => {
   const total = studiesResults?.data?.hits?.total || 0;
 
+  const labelByFieldKeyMap: Record<string, string> = {
```
adipaul1981, line 23: Not sure about the strategy here. There is already the Display Name that you can use from querying the arranger mapping. (extended mapping). See `studiesMappingResults ` in studies.tsx. Maybe use this to build the dictionary?

adipaul1981 (changes requested):

evans-g-crsj (approved):
Code is a;; good for me. But for the rest (Adrian's comment) I can't judge.

src/pages/studies/StudyPageContainer.tsx
celinepelletier: Thanks for your comment, I was not aware of this mapping, I made changes to use it :)

Issue #3062, opened 2021-04-09
Participant set's name suggestion doesn't work anymore


vferretti:
Small regression: No set name is suggested by default when we save a new participant set in the cohort builder, 
![image.png](https://images.zenhubusercontent.com/5bbcbdee387e41362246faa0/c22b2dfd-95dd-4252-a42a-83112859952f)

PR #3171, opened 2021-04-28
fix: #3062 saveset default name

adipaul1981:
image

```
Test Suites: 27 passed, 27 total
Tests:       155 passed, 155 total
Snapshots:   1 passed, 1 total
Time:        7.648s
Ran all test suites.
```

### DIFF
--- a/src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
+++ b/src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
@@ -198,7 +198,7 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
     const formDefaultValue =
       saveSetActionType === SaveSetActionsTypes.CREATE
         ? { nameSet: defaultName !== SET_DEFAULT_NAME ? defaultName : '' }
-        : { nameSet: setToRename?.name || '' };
+        : { nameSet: setToRename?.name ?? '' };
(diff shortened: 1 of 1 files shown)


evans-g-crsj (approved):

src/components/CohortBuilder/ParticipantsTableView/SaveSetModal.tsx
```diff
@@ -196,7 +195,11 @@ const SaveSetModal: FunctionComponent<Props> = (props) => {
   useEffect(() => {
     const defaultName = filtersToName({ filters: sqon });
     setLoadingDefaultTagName(true);
-    setDefaultTagName(defaultName !== SET_DEFAULT_NAME ? defaultName : '');
+    const formDefaultValue =
+      saveSetActionType === SaveSetActionsTypes.CREATE
+        ? { nameSet: defaultName !== SET_DEFAULT_NAME ? defaultName : '' }
+        : { nameSet: setToRename?.name || '' };
```
celinepelletier, line 201: maybe use `??` instead of `||`

`??` is a logical operator that returns its right-hand side operand when its left-hand side operand is null or undefined, and otherwise returns its left-hand side operand. Whereas `||` returns its right-hand side operand when its left-hand side is false (so not only null or undefined)

celinepelletier (approved):

Issue #3165, opened 2021-04-27
Variant page : Add tooltip to participants column in KF Studies table


luclemo:
Add a tooltip to help users understand the discrepancy between the number displayed in the link and the value returned.

**Message:**
`Due to participant confidentiality, links may return a smaller number than displayed.`

**Mockup:**
https://www.figma.com/file/xelYw9knAeZJomJ2lh4xfa/KF-Variant-Page?node-id=1280%3A23857

![Pasted_Image_2021-04-27__5_15_PM.png](https://images.zenhubusercontent.com/5dadb279f8e0d70001993615/2e8241c8-f952-4096-b34d-c77cf144cb5a)

PR #3206, opened 2021-05-05
feat(FreqTab): #3165 add participants tooltip

evans-g-crsj:
#3165 
participantsLinksWith

participantsLinksWithout

```
Test Suites: 25 passed, 25 total
Tests:       142 passed, 142 total
Snapshots:   0 total
Time:        5.19 s
Ran all test suites.
```
**Note:** I had to bypass eslint again since it doesn't like variables in scss. However, I made sure to correct all that I could.

### DIFF
--- a/src/pages/variantEntity/TabFrequencies.tsx
+++ b/src/pages/variantEntity/TabFrequencies.tsx
@@ -2,8 +2,9 @@
 import React from 'react';
 import { connect, ConnectedProps } from 'react-redux';
 import { Link } from 'react-router-dom';
+import { InfoCircleOutlined } from '@ant-design/icons';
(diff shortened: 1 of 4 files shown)


celinepelletier (approved):

francisl (changes requested):
What is different from those tooltip compared to the default theme? geekblue-7 seem to be the default theme color https://www.figma.com/file/MKWJJXL7sz4poiIv5bKNtz/KF-UI-Kit-ant.design-v4-update?node-id=0%3A1617

Why do you need to export the color, is the antd tooltip component className support broken?

does the override works? https://github.com/ant-design/ant-design/blob/master/components/tooltip/style/index.less#L61

PR #3207, opened 2021-05-05
incorporate code review comments

evans-g-crsj:


src/style/themes/default/antd/tootltip.less
```diff
@@ -0,0 +1,5 @@
+.@{tooltip-prefix-cls} {
+  &-inner {
+    color: @tooltip-color;
+  }
+}
```
francisl, line 5: why is this necessary? since you already settle value of the variable, the theme should pick it up.

evans-g-crsj: good point!

francisl (approved):

Issue #3131, opened 2021-04-20
Update Filters Style 


francisl:
Facets panel: implement styling as defined by Lucas in the spec

francisl:
@luclemo @vferretti please provide more information. The filters were build against the figma template. Did something changes after the time we push our code? does something excape the eyes of a simple developer :-) ?

luclemo:
@francisl I think what @vferretti is referring to is the spacing and text colors/weights of the facet components (view screenshot). Nothing has changed from the [facet mockups file](https://www.figma.com/file/cnK68RMYFGoLbgsnCYvoo6/FerlabUI?node-id=659%3A21875). 

Let's speak first because I have a small change I would like to include with the size of the sider and the text size of the checkbox labels.

![image](https://user-images.githubusercontent.com/1451515/115708738-5e47dd80-a33e-11eb-8e96-ba567d83440e.png)

luclemo:
### Tasks

@francisl Following our discussion today we are looking to:

- [ ] Expand the width of the sidebar with a **min/max** of **320px and 420px** ([see coded example here](https://codepen.io/luclemo/pen/7d8bf028f22cc92ab4eb53144186ce15))
- [ ] Revert the **checkbox labels** to the default UI font size (14px)
- [ ] Ensure proper alignment of checkbox element and number tag (see specs in mockup link below)
- [ ] Adjust colors + font weights of text elements as per the UI kit

**Note:** these changes will apply to all FerLab facet components.

### Mockups

- Mockup example of full sidebar layout: [Studies page mockup](https://www.figma.com/file/IelL4AHNbrBxDA94zSOxJt/?node-id=123%3A8229)

PR #3230, opened 2021-05-07
fix: #3131 update sidebar filter wrap colors and size

francisl:
#3131 
Screen Shot 2021-05-07 at 5 11 09 PM

### DIFF
--- a/src/pages/studies/Sidebar.module.scss
+++ b/src/pages/studies/Sidebar.module.scss
@@ -64,7 +64,3 @@
     border: none;
   }
 }
-
-:export {
-  width: 420;
-}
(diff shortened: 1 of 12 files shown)


src/pages/studies/Sidebar.module.scss
```diff
@@ -44,9 +59,12 @@
...
   & :global(.ant-collapse),
-  :global(.ant-collapse-content),
   :global(.ant-collapse-item) {
     background-color: $gray-1;
     border: none;
   }
 }
+
+:export {
+  width: 420;
```
evans-g-crsj, line 69: Here, it's clear that width is related to the sidebar. However, I doubt it will be as obvious of a name when using this variable in the calling file.

evans-g-crsj (approved):

src/pages/studies/Sidebar.module.scss
francisl: in the refactoring I removed the need for it. So i'll delete it.  good catch

celinepelletier (approved):

PR #3241, opened 2021-05-11
Fix/3238 study graph display study short name onhover

adipaul1981:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "kf-portal-ui",
-  "version": "3.0.2",
+  "version": "3.0.3",
   "private": true,
   "dependencies": {
     "@apollo/client": "^3.3.11",
--- a/src/components/CohortBuilder/Summary/Cards/StudiesChart.js
+++ b/src/components/CohortBuilder/Summary/Cards/StudiesChart.js
@@ -21,6 +21,11 @@ const sortDescParticipant = (a, b) => {
   return aTotal - bTotal;
 };
 
+const bucketTooltipExtract = (b) => ({
+  label: b.key,
+  tooltip: b.top_hits.study.short_name,
+});
+
 export const studiesQuery = (sqon) => ({
   query: gql`
     query($sqon: JSON) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "kf-portal-ui",
-  "version": "3.0.2",
+  "version": "3.0.3",
```
evans-g-crsj, line 3: make sure that all the team says go for the version bumping since it's a hotfix (I don't know if we should or not change the version).

src/components/CohortBuilder/Summary/Cards/StudiesChart.js
```diff
@@ -66,11 +68,23 @@ export const studiesQuery = (sqon) => ({
...
       (pB) => (studyLabelToCounts[pB.key] = { probands: pB.doc_count, familyMembers: 0 }),
     );
 
     const familyMembersBuckets = participants.familyMembers.study__code.buckets;
+
+    const studyDictionaryFamily = familyMembersBuckets.map((b) => ({
+      label: b.key,
+      tooltip: b.top_hits.study.short_name,
+    }));
+
```
evans-g-crsj, line 87: the callback in map since identical. I think it would be better to extract it.

src/components/CohortBuilder/Summary/Cards/StudiesChart.js
```diff
@@ -82,36 +96,16 @@ export const studiesQuery = (sqon) => ({
...
-    return Object.entries(studyLabelToCounts).reduce((accumulator, [label, counts]) => {
-      const { familyMembers, probands } = counts;
-      return [...accumulator, { label, familyMembers, probands, id: toKebabCase(label) }];
-    }, []);
+    return {
+      data: Object.entries(studyLabelToCounts).reduce((accumulator, [label, counts]) => {
+        const { familyMembers, probands } = counts;
+        return [...accumulator, { label, familyMembers, probands, id: toKebabCase(label) }];
+      }, []),
+      dictionary: [...studyDictionaryProbands, ...studyDictionaryFamily],
```
evans-g-crsj, line 104: It can be viewed as a dictionary indeed. However, I don't think the name is meaningful. I would name it differently studyToCounts or something like that.

evans-g-crsj (changes requested):

package.json
francisl: “Given a version number MAJOR.MINOR.PATCH, increment the:
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.”

https://semver.org

Patch est ok pour des hotfixs

evans-g-crsj (approved):

Issue #3247, opened 2021-05-12
Impossible to delete a virtual study


denis-crsj:
Impossible to delete a Virtual Study. I tried deleting the study from the cohort builder and also from the Dasboard card and both ways don't work. A modal appears with an error message stating that an error occurred and that the deletion did not work.

**Usersnap Feedback**
![image](https://user-images.githubusercontent.com/64538118/118005465-692bd780-b318-11eb-9ade-d04fdd284524.png)

![image](https://user-images.githubusercontent.com/64538118/118005413-5c0ee880-b318-11eb-81b2-fb2194ca7912.png)

PR #3253, opened 2021-05-12
fix: #3247 fix delete request

celinepelletier:
Since latest version of axios, delete, get, options and head don't support body argument, instead data should be passed by config argument.

Issue: https://github.com/axios/axios/issues/897
Documentation: https://github.com/axios/axios#request-method-aliases

So I update api.js to pass data in different way depending on method

### DIFF
--- a/src/services/api.js
+++ b/src/services/api.js
@@ -18,7 +18,8 @@ export const initializeApi = ({
   arrangerRoot = arrangerApiRoot,
 }) => {
   const uri = url || urlJoin(arrangerRoot, endpoint);
-  return sendRequest(defaultHeaders, method, body, headers, uri)
+  const methodLowerCase = (method || '').toLowerCase();
+  return sendRequest(defaultHeaders, methodLowerCase, body, headers, uri)
     .then((response) => response.data)
     .catch((err) => {
       const { response } = err;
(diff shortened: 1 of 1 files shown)


src/services/api.js
```diff
@@ -35,14 +29,41 @@ export const initializeApi = ({
...
+const sendRequest = (defaultHeaders, method, body, headers, uri) => {
+  switch (method.toLowerCase()) {
+    case 'post':
+    case 'put':
+    case 'patch':
+      return ajax[method.toLowerCase()](uri, body, {
+        headers: {
+          'Content-Type': 'application/json',
+          ...(defaultHeaders || {}),
+          ...headers,
```
evans-g-crsj, line 41: should that be in a const since it's used at two places?

src/services/api.js
```diff
@@ -35,14 +29,41 @@ export const initializeApi = ({
     });
 };
 
+const sendRequest = (defaultHeaders, method, body, headers, uri) => {
+  switch (method.toLowerCase()) {
```
evans-g-crsj, line 33: maybe method.toLowerCase() could be protected and used in a const. For example: const methodLowerCase = (method || '').toLowerCase()

src/services/api.js
```diff
@@ -35,14 +29,41 @@ export const initializeApi = ({
...
+    case 'options':
+      return ajax[method.toLowerCase()](uri, {
+        headers: {
+          'Content-Type': 'application/json',
+          ...(defaultHeaders || {}),
+          ...headers,
+        },
+        data: body,
+      });
+  }
```
evans-g-crsj, line 56: I don't see a default trap

evans-g-crsj (approved):

Issue #3244, opened 2021-05-12
build: decrease vulnerabilities from dependencies as much as possible


evans-g-crsj:
from npm audit : `found 676 vulnerabilities (288 low, 168 moderate, 220 high) in 3392 scanned packages`

celinepelletier:
TODO : 
<img width="1290" alt="Capture d’écran, le 2021-05-12 à 10 25 35" src="https://user-images.githubusercontent.com/82821620/117991935-7fcc3180-b30c-11eb-8ed1-de0fe4b3f8cd.png">

celinepelletier:
<img width="304" alt="Capture d’écran, le 2021-05-18 à 09 52 54" src="https://user-images.githubusercontent.com/82821620/118663662-db8c3400-b7be-11eb-85a4-e3ecfdfb74f8.png">

celinepelletier:
Also fix babel-polyfill deprecated warning message

PR #3270, opened 2021-05-17
fix: #3244 fix warning message about babel polyfill deprecated

celinepelletier:
I needed to ignore eslint import order alerts since if I change the order of the first 5 imports, it breaks the UI

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
-import 'babel-polyfill';
+/* eslint-disable simple-import-sort/imports */
+import 'react-app-polyfill/stable';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1,4 +1,5 @@
-import 'babel-polyfill';
+/* eslint-disable simple-import-sort/imports */
```
evans-g-crsj, line 1: maybe add a comment that explains why imports have to placed in that particular order?

evans-g-crsj (approved):

Issue #3275, opened 2021-05-19
fix(profile): fix form


evans-g-crsj:
make sure that a user can correctly use these fields:  _institution_, _institutional email_ and _title_

PR #3276, opened 2021-05-19
fix(profile): #3275 make sure form fields work fine and fix find me

evans-g-crsj:
image


src/components/UserProfile/ContactReadInformation.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+
+const generateContactValueStyle = (info: FieldValue) => (info ? '' : 'contact-info-value');
+
+const convertValueToEmailLinkIfNeeded = (value: FieldValue, isEmail: boolean) => {
+  if (isEmail && value) {
+    return <a href={`mailto:${value}`}>{value}</a>;
+  }
+  return value;
+};
```
francisl, line 23:
```
const convertValueToEmailLinkIfNeeded = (value: FieldValue, isEmail: boolean) => (isEmail && value) 
   ? <a href={`mailto:${value}`}>{value}</a>
   : value;
```

src/components/UserProfile/utils.js
```diff
@@ -106,88 +108,88 @@ export const userProfileBackground = (
...
-      </Fragment>
-    ),
-  });
+  loading: isProfileUpdating,
+  title: <Text className={'header-title'}>{title}</Text>,
+  className: 'card-container',
+  headStyle: style.cardHeadStyle,
+  bodyStyle: style.cardBodyStyleWhenEditing,
+  extra: (
+    <Fragment>
```
francisl, line 163: ```<></>```

francisl (changes requested):
Some comments.

change what you think that make sense. But the <fragment> should be remove for consistency.

src/components/UserProfile/ContactReadOnly.tsx
```diff
@@ -0,0 +1,158 @@
...
+    return false;
+  }
+  const role = extractRoleFromProfile(data);
+  let fieldsToCheckFor = COMMON_PROPERTIES;
+  if (role === 'research') {
+    fieldsToCheckFor = fieldsToCheckFor.concat(RESEARCHER_SPECIFIC_PROPERTIES);
+  } else if (role === 'community') {
+    fieldsToCheckFor = fieldsToCheckFor.concat(COMMUNITY_SPECIFIC_PROPERTIES);
+  }
+  return fieldsToCheckFor.some((f) => data[f as keyof LoggedInUser]);
```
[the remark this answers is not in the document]
evans-g-crsj: In this case, the goal is to verify if the data received are empty. Therefore, I'm not sure about the benefits and will not modify the code. However, you have a good point and I think it would be nice to construct a User object with utility functions (some that would be related to the `role`) that would simply some logic used a crossed the portal.

PR #3329, opened 2021-06-10
feat: #3328 add nginx in docker

celinepelletier:
I'm not able to test for Linux, so if one of you can test and let me know if something doesn't work.


docker/nginx/prod.crt
```diff
@@ -0,0 +1,22 @@
+-----BEGIN CERTIFICATE-----
```
adipaul1981, line 1: this file should be called `prd.crt` (see Dockerfile)

docker/nginx/prod.key
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
adipaul1981, line 1: this file should be called `prd.key` (see Dockerfile)

adipaul1981 (approved):

oliviercperrier (approved):

Issue #3344, opened 2021-06-22
 Diagnosis (source text) show incorrect `No Data`


adipaul1981:
**Bug description**
Many Diagnosis (source text) show are `No Data`

**Steps To Reproduce**
1. Go explore data page
2. Click on Clinical -> Diagnosis (Source Text)
3. Lots of No Data count

https://kidsfirst-team.slack.com/archives/C01G7Q29H5M/p1624052089016400

**Expected behavior**
Not show inaccurate `No Data`

**Screenshots**
![image](https://user-images.githubusercontent.com/29788342/122963005-e30da280-d353-11eb-87f8-e9e897724e24.png)

PR #3348, opened 2021-06-23
fix: #3344 diagnosis fields show incorrect No data count

adipaul1981:


celinepelletier:
ton package lock est en version 1, on est passé en version 2 maintenant, il faut que tu vérifies que tu es bien en : 
 node: >=15.14.0 et npm: >=7.7.6

celinepelletier (changes requested):
package-lock should have any changes since you didn't add/update/delete any dependancy. Check your node and npm versions, package-lock should be in version 2

celinepelletier (approved):

PR #3355, opened 2021-06-30
#3352 remove saved queries from freactal

evans-g-crsj:
image


src/store/SavedQueriesTypes.ts
```diff
@@ -0,0 +1,110 @@
+import { ThunkDispatch } from 'redux-thunk';
+
+import { RootState } from './rootState';
+import { VirtualStudy } from './virtualStudiesTypes';
+
+type FilesQueryContent = {
+  Families: number;
```
celinepelletier, line 7: why do theses properties start with an uppercase letter ?

src/store/actionCreators/SavedQueries.ts
```diff
@@ -0,0 +1,100 @@
...
+const successDeletingSavedQueryAction = (
+  queryId: string,
+  queryType: QueryType,
+): SavedQueriesActionTypes => ({
+  type: SavedQueriesActions.successDeletingSavedQuery,
+  queryId,
+  queryType,
+});
+
+const failureDeletingQuery = (queryId: string, error: Error | null): SavedQueriesActionTypes => ({
```
celinepelletier, line 47: I don't see a case where error would be null

src/store/reducers/SavedQueries.ts
```diff
@@ -0,0 +1,60 @@
...
+    case SavedQueriesActions.failureFetchingAllSavedQueries: {
+      return { ...state, errorFetchAllQueries: action.error };
+    }
+    case SavedQueriesActions.requestDeleteSavedQuery: {
+      const queryId = action.queryId;
+
+      const queryIdToStatusShallowCopy = { ...state.queryIdToStatus };
+      queryIdToStatusShallowCopy[queryId] = SavedQueriesStatuses.deleting;
+
+      return { ...state, queryIdToStatus: queryIdToStatusShallowCopy };
```
celinepelletier, line 33:
```
return { ...state, queryIdToStatus: { ...state.queryIdToStatus, [queryId]:  SavedQueriesStatuses.deleting} };
```
would also work i think

src/store/reducers/SavedQueries.ts
```diff
@@ -0,0 +1,60 @@
...
+        ...state,
+        queries: { ...queriesShallowCopy, [queryType]: filteredQueriesForType },
+        queryIdToStatus: queryIdToStatusShallowCopy,
+      };
+    }
+    case SavedQueriesActions.failureDeletingSavedQuery: {
+      const queryId = action.queryId;
+      const queryIdToStatusShallowCopy = { ...state.queryIdToStatus };
+      queryIdToStatusShallowCopy[queryId] = SavedQueriesStatuses.error;
+      return { ...state, queryIdToStatus: queryIdToStatusShallowCopy };
```
celinepelletier, line 55:
```
return { ...state, queryIdToStatus: { ...state.queryIdToStatus, [queryId]:  SavedQueriesStatuses.error} };
```

src/store/reducers/SavedQueries.ts
```diff
@@ -0,0 +1,60 @@
...
+    case SavedQueriesActions.requestDeleteSavedQuery: {
+      const queryId = action.queryId;
+
+      const queryIdToStatusShallowCopy = { ...state.queryIdToStatus };
+      queryIdToStatusShallowCopy[queryId] = SavedQueriesStatuses.deleting;
+
+      return { ...state, queryIdToStatus: queryIdToStatusShallowCopy };
+    }
+    case SavedQueriesActions.successDeletingSavedQuery: {
+      const queryId = action.queryId;
```
celinepelletier, line 36: I usually avoid to have complex logic in reducer cases, I would probably add some utils function to remove elem from a list/ a map (or if we have kind libs like lodash, i would use them), and just call them here (moreover, they could be useful in others places of the app)

celinepelletier (approved):
Some comments but nothing blocking.
Very nice job 💯

src/store/SavedQueriesTypes.ts
evans-g-crsj: I had the same reaction. I just don't know - that's how the data is received.

src/store/actionCreators/SavedQueries.ts
evans-g-crsj: Good point, I'll change that.

src/store/reducers/SavedQueries.ts
evans-g-crsj: Good point. I'll change that.

src/store/reducers/SavedQueries.ts
evans-g-crsj: Yes, it would but I have mixed feelings about it. It's a one-liner, but seems a bit more complex than it needs to. Maybe the other reviewers can give their take on it.

src/store/reducers/SavedQueries.ts
evans-g-crsj: same comment as : https://github.com/kids-first/kf-portal-ui/pull/3355#discussion_r661815537

src/store/reducers/SavedQueries.ts
evans-g-crsj: I retract my last comment. It is indeed better.

adipaul1981 (approved):

PR #3358, opened 2021-07-07
#3346 duplicate freactal fences provider into redux

evans-g-crsj:


src/hooks/useFenceConnections.ts
```diff
@@ -0,0 +1,40 @@
...
+} from 'store/selectors/fenceConnections';
+
+type Output = {
+  isFetchingAllFenceConnections: boolean;
+  fenceConnections: FenceConnections;
+  fencesAllAcls: FencesAllConcatenatedAcls;
+};
+
+const useFenceConnections = (api: Api, fences: string[]): Output => {
+  const fenceConnections = useSelector((state: RootState) => selectFenceConnections(state));
```
evans-g-crsj, line 24: have to make sure fence connections exist before fetching studies

src/hooks/useFenceStudies.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const useFenceStudies = (api: Api): Output => {
+  const dispatch: DispatchFenceStudies = useDispatch();
+
+  const isFetchingAllFenceStudies = useSelector((state: RootState) =>
+    selectIsFetchingAllFenceStudies(state),
+  );
+  const fenceStudies = useSelector((state: RootState) => selectFenceStudies(state));
+  const fenceConnections = useSelector((state: RootState) => selectFenceConnections(state));
```
evans-g-crsj, line 30: have to make sure fence connections exist before fetching studies

src/services/fenceStudies.js
```diff
@@ -0,0 +1,99 @@
+import keys from 'lodash/keys';
+
+import { graphql } from 'services/arranger';
+
+export const getAuthStudiesIdAndCount = async (api, fence, userAcl) =>
```
evans-g-crsj, line 5: Copied...Will do some cleanup later on.

src/store/actionCreators/fenceConnections.ts
```diff
@@ -0,0 +1,83 @@
...
+
+const fetchFencesConnections = (
+  api: Api,
+  fenceName: FenceName,
+): ThunkAction<void, RootState, null, FenceConnectionsActionTypes> => async (dispatch) => {
+  try {
+    const connection = await getFenceUser(api, fenceName);
+    dispatch(addFenceConnection(fenceName, connection));
+  } catch (error) {
+    console.error(`Error fetching fence connection for '${fenceName}': ${error}`);
```
evans-g-crsj, line 44: In current code (freactal), there are no error management. However, I prepared the ground for the future if we need to manage it.

adipaul1981 (approved):

celinepelletier (approved):

adipaul1981 (approved):

adipaul1981 (approved):

src/store/actionCreators/Test/fenceConnections.test.ts
```diff
@@ -61,8 +53,11 @@ describe('Fence Connections actions', () => {
   });
 
   it('should manage acls adequately', () => {
+    // @ts-ignore
```
celinepelletier, line 56: why do you need to ignore ts here ?

celinepelletier (approved):

src/store/actionCreators/Test/fenceConnections.test.ts
francisl: `ts-ignore` requires an detailed explanation and justification

evans-g-crsj: @celinepelletier remember the Nullable? That's why I need a ts-ignore, because, as received from the network, some fields can be null. However in the Connection Type def I didn't say that a field can be null.

image

celinepelletier: is it a real use case that API returns null values ?

src/components/UserDashboard/index.js
```diff
@@ -28,32 +28,38 @@ import {
...
+)(({ state: { loggedInUser }, api }) => {
+  if (!loggedInUser) {
+    return (
+      <div className={userDashboardContainer}>
+        <div className={userDashboardContent}></div>
+        <Spinner />
+      </div>
+    );
+  }
+  return (
```
francisl, line 40:
```
=> !loggedInUser ? 
   <div>...<div> :
   <div>else</div>;
```
?

src/components/UserProfile/Integration.js
```diff
@@ -52,12 +58,19 @@ const trackFenceAction = async ({ fence, fenceDetails, category, action, label }
...
   setConnecting(true);
   try {
     await deleteFenceTokens(api, fence);
-    await effects.setIntegrationToken(fence, null);
-    await effects.removeFenceConnection(fence);
+    removeFenceConnection(fence);
+    removeFenceStudies(fence);
     await trackFenceAction({
       fence,
       fenceDetails: '',
```
francisl, line 76: try with no `catch` not considered a good pratice. I know it's not from that PR, but worth looking into

src/store/actionCreators/Test/fenceConnections.test.ts
evans-g-crsj: When I log the data received from a fence - I get something similar to the mock object (a lot of null values in my case).

However, I can just remove those fields in the mock object.

francisl (commented):
Looks good, small comments

src/components/UserProfile/Integration.js
evans-g-crsj: There is a `catch` clause. It is just not visible in the snippet above.

src/components/UserDashboard/index.js
evans-g-crsj: ok! Will change that.

src/store/actionCreators/Test/fenceConnections.test.ts
celinepelletier: ok so you could declare your property as : 
`display_name?: string | null`
It should do the job

evans-g-crsj: it would. That's why I used Nullable last time as a kinda abbreviation. For the tests I just removed the null fields from the mock object.

francisl (approved):
:shipit:

PR #3361, opened 2021-07-12
#skfp-24 make sure saved queries display correctly

evans-g-crsj:
image

### DIFF
--- a/src/components/UserDashboard/SavedQueries/CohortTab.tsx
+++ b/src/components/UserDashboard/SavedQueries/CohortTab.tsx
@@ -22,6 +22,11 @@ type Props = {
   queryIdToStatus: SavedQueriesIdToStatus;
 };
 
+const MAX_N_OF_CHARS_FOR_VS_DESCRIPTION = 140;
+
+const truncateWithEllipsis = (description: string) =>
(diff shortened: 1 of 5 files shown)


src/components/UserDashboard/SavedQueries/CohortTab.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  return (
+    <SavedQueriesTab
+      items={queries.cohort}
+      queryIdToStatus={queryIdToStatus}
+      makeDescription={(item) => {
+        const vs = item as VirtualStudyPlusId;
+        return (
+          <>
+            {vs.description.length >= 140 ? `${vs.description.slice(0, 140)}...` : vs.description}
```
celinepelletier, line 36: a detail but i would put 140 in a variable to avoid magic number :)

celinepelletier (approved):

adipaul1981 (approved):

src/components/UserDashboard/SavedQueries/CohortTab.tsx
evans-g-crsj: yes, good point.

PR #3362, opened 2021-07-13
#SKFP-37 fix icons

evans-g-crsj:


src/style/dist/themes/default/antd.css
```diff
@@ -25052,3 +25052,6 @@ h3.ant-typography {
 .ant-collapse-header {
   font-weight: 600;
 }
+.ant-list-item-meta-title > a {
+  color: #a6278f;
```
celinepelletier, line 25056: is it a new color ? not in existing colors ?

src/components/CohortBuilder/Categories/index.js
```diff
@@ -95,7 +97,7 @@ class Categories extends React.Component {
           color={theme.studyRed}
           anchorId={'anchor-study'}
         >
-          <StudyIcon fill={theme.studyRed} />
+          <StudyIcon fill={theme.studyRed} width="14px" height="17px" />
```
celinepelletier, line 100: is it the issue Adrian was talking about to pass fill and classname ?

package-lock.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "kf-portal-ui",
-  "version": "3.1.3",
+  "version": "3.2.0",
```
celinepelletier, line 3: I would bump version in a separate commit which  would include bump + release note
The release tag will be on this specific commit 
Just to clarify history :)

celinepelletier (approved):

src/style/dist/themes/default/antd.css
francisl: ⚠️ should be in the color theme.

src/components/CohortBuilder/Categories/index.js
francisl: ☝️

francisl (changes requested):
hardcoded need fixes.

src/components/CohortBuilder/Categories/index.js
evans-g-crsj: yes

src/style/dist/themes/default/antd.css
evans-g-crsj: This is the final css file. There is no use for using variables here.

src/components/CohortBuilder/Categories/index.js
evans-g-crsj: the issue was that no `fill` prop was passed down to the icon. We (me and Adrian) added the prop. The `width="14px" height="17px" ` (similarly to the rest of the icons in the file) was added so that the icon be the same size as the other icons in this component (pls, see the whole file).

adipaul1981: the icons in the portal are a complete mess. they are all different, and don't work in the same way. Anyway, its such a tedious job that nobody wants to tackle it. :)

adipaul1981 (approved):

src/style/dist/themes/default/antd.css
celinepelletier: oh yes you're right sorry about this

src/components/CohortBuilder/Categories/index.js
celinepelletier: if there is no way to fix it, i'm confortable to let it like this

celinepelletier:
Looks good to me

src/components/CohortBuilder/Categories/index.js
francisl: But the icon is in `src/icons`, why not 'normalize' our icons there instead of a per use basis.

francisl: @adipaul1981 🐥 🥚

francisl (approved):

PR #3369, opened 2021-07-22
fix(reduxdevtool): #SKFP-56 remove redux dev tool in prod

oliviercperrier:


src/store/index.js
```diff
@@ -1,6 +1,6 @@
-import { createStore, applyMiddleware } from 'redux';
-import thunk from 'redux-thunk';
-import { composeWithDevTools } from 'redux-devtools-extension/logOnlyInProduction';
```
evans-g-crsj, line 3: I think that it could do the job `'redux-devtools-extension/developmentOnly'`

evans-g-crsj (changes requested):

evans-g-crsj (approved):

celinepelletier (approved):

PR #3377, opened 2021-08-16
Fences behaviour when logout

evans-g-crsj:
## Description

make sure that connection to fences are not lost when a user logout but connections to fences are still on.

## Validation

- [ ] Code Approved
- [x] Test Coverage
- [ ] QA Done
- [x] Design/UI Approved from design

(connect successfully to fences, then logout then login => settings should be unchanged)
image

## QA

connect successfully to fences, then logout then login 
...

## Mention

@ QA, Design ...

### DIFF
--- a/src/components/Login/Login.js
+++ b/src/components/Login/Login.js
@@ -92,7 +92,8 @@ const Component = (props) => {
     if ((response || {}).status === 200) {
       const rawJwt = response.data;
       manageUserTokenWithLoader(rawJwt, provider, () => {
-        setIsHandlingToken(false);
+        // Note: no need to use setIsHandlingToken(false);
+        // When changing page this component (login) will unmount.
         history.push(ROUTES.dashboard);
       });
       await trackUserSignIn(provider);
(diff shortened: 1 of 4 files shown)


src/components/Login/Login.js
```diff
@@ -92,7 +92,8 @@ const Component = (props) => {
     if ((response || {}).status === 200) {
       const rawJwt = response.data;
       manageUserTokenWithLoader(rawJwt, provider, () => {
-        setIsHandlingToken(false);
+        // Note: no need to use setIsHandlingToken(false);
```
evans-g-crsj, line 95: I had warnings because I set a state on an unmounted component. So, I corrected that while at it.

src/store/actionCreators/user.ts
```diff
@@ -205,7 +201,7 @@ export const fetchUserFromJwt = (validDecodedJwt: DecodedJwt): ThunkActionUser =
     dispatch(receiveUser(profile));
   } catch (error) {
     console.error(error);
-    dispatch(revertAcceptedTermsThenLogoutCleanly());
+    dispatch(cleanlyLogout());
```
evans-g-crsj, line 204: enforcing to remove accepted terms from the front-end lead to errors in some cases (e.g unauthorized call to persona). Therefore, I rather not enforce it too much now since we will take care of this with keycloak in the near future.

oliviercperrier (approved):

PR #3401, opened 2021-09-16
fix: #SKFP-140 hotfix login with orcid always redirect to login page

celinepelletier:
master to dev after hotfix SKFP-140


francisl (approved):
:shipit:

PR #3435, opened 2021-10-19
Skfp 248

oliviercperrier:
# FIX

- closes #SKFP-248


celinepelletier:
Your package-lock is generated with version 1 instead of 2. Make sure you use node 15+ and npm 7+

PR #3509, opened 2022-08-31
:lipstick: feat(study): SKFP-353 add study page

lflangis:
# FEATURE

- closes [353](https://d3b.atlassian.net/browse/SKFP-353)

## Description

Goal: Implement the Studies page as per the specifications in the Notion page below. 
https://www.notion.so/ferlab/Study-Page-Remodel-3005c26a269b4991a5fffdedb1bba3eb

## Screenshot

https://user-images.githubusercontent.com/65532894/187693333-e22b8625-223c-4a77-8fd4-5e3b50857fd1.mp4

### DIFF
--- a/src/utils/helper.ts
+++ b/src/utils/helper.ts
@@ -40,3 +40,8 @@ export const toKebabCase = (str: string) => {
   const match: string[] = (str && str.match(KEBAB_REGEX)) || [];
   return match.map((x: string) => x.toLowerCase()).join('-');
 };
+
+export const toPascalCase = (str: string) =>
+  str.replace(/(\w)(\w*)/g, function (g0, g1, g2) {
+    return g1.toUpperCase() + g2.toLowerCase();
+  });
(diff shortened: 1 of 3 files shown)


src/utils/helper.ts
```diff
@@ -40,3 +40,8 @@ export const toKebabCase = (str: string) => {
   const match: string[] = (str && str.match(KEBAB_REGEX)) || [];
   return match.map((x: string) => x.toLowerCase()).join('-');
 };
+
+export const toPascalCase = (str: string) =>
```
evans-g-crsj, line 44: nice utils but since we have lodash installed I wonder if we could just do:

`startCase(str.toLower());`

src/views/Studies/components/PageContent/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+    sqon: resolvedSqon,
+  });
+
+  useEffect(() => {
+    setQueryConfig({
+      ...queryConfig,
+      pageIndex: DEFAULT_PAGE_INDEX,
+    });
+    // eslint-disable-next-line
+  }, [JSON.stringify(activeQuery)]);
```
evans-g-crsj, line 60: is there a particular field that we could check (an id for instance). JSON.stringnify can lead to subtle bugs.

evans-g-crsj (approved):

PR #3242, opened 2021-05-12
3211 fix pagination style to reflect theme

francisl:
Screen Shot 2021-05-12 at 9 29 33 AM

PR #3517, opened 2022-09-21
:lipstick: feat(zeppelin): SKFP-452 add zeppelin widget

lflangis:
# Feature

- closes #3242

## Description
Set up the Apache Zeppelin Widget as it is in KF. With successful connection, it will redirect to the VWB

Design
image

Now

Screenshot_20220921_154918
Screenshot_20220921_155008

https://user-images.githubusercontent.com/65532894/191597868-adccc919-7240-4357-ae26-8413f6a4078c.mp4

https://user-images.githubusercontent.com/65532894/191597919-57aa613d-7697-4b0b-a2ab-4fe24a8049be.mp4

https://user-images.githubusercontent.com/65532894/191597931-aa07083a-f0de-45ad-b68c-15545decff4f.mp4


src/components/uiKit/Fences/Modal/index.tsx
```diff
@@ -17,16 +17,21 @@ const iconSize = {
...
+  const { modalConnectionParams, connectionStatus, fencesConnectError, loadingFences } =
     useFenceConnection();
 
   const isDcfConnected = connectionStatus[FENCE_NAMES.dcf] === FENCE_CONNECTION_STATUSES.connected;
   const isGen3Connected =
     connectionStatus[FENCE_NAMES.gen3] === FENCE_CONNECTION_STATUSES.connected;
 
-  const onClose = () => dispatch(fenceConnectionActions.toggleConnectionModal(false));
+  const onClose = () => {
+    if (modalConnectionParams.onClose) {
```
evans-g-crsj, line 30: 1. suggestion (optional): `modalConnectionParams.onClose && modalConnectionParams.onClose()`
2. You should not have to explicitly pass down `undefined` . However, if the Types are a little bit messy, forget about it.

src/helpers/notebook.ts
```diff
@@ -0,0 +1,14 @@
+import { NotebookApiStatus } from 'services/api/notebook/model';
```
evans-g-crsj, line 1: very good

src/helpers/notebook.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const isNotebookStatusLaunched = (status: NotebookApiStatus) =>
+  [
+    NotebookApiStatus.createComplete,
+    NotebookApiStatus.stopped,
+    NotebookApiStatus.rollback,
+  ].includes(status);
+
+export const isNotebookStatusInProgress = (status: NotebookApiStatus) =>
+  status === NotebookApiStatus.deleteInProgress || status === NotebookApiStatus.createInProgress;
```
evans-g-crsj, line 11: I think you could use `includes` similarly to the function right above.

src/services/api/notebook/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const start = () => {
+  return sendRequest<TNotebookApiResponse>({
+    method: 'POST',
+    url: NOTEBOOK_API_URL,
+  });
+};
+
+const get = () => {
+  return sendRequest<any>({
```
evans-g-crsj, line 15: if you can (easily) change the type

src/services/api/notebook/index.ts
```diff
@@ -0,0 +1,31 @@
+import EnvironmentVariables from 'helpers/EnvVariables';
+import { sendRequest } from 'services/api';
+import { TNotebookApiResponse } from './model';
+
+export const NOTEBOOK_API_URL = `${EnvironmentVariables.configFor('VARIANT_CLUSTER_API')}`;
+
+const start = () => {
+  return sendRequest<TNotebookApiResponse>({
+    method: 'POST',
```
evans-g-crsj, line 9: maybe use an enum for http methods.

src/store/notebook/slice.ts
```diff
@@ -0,0 +1,88 @@
...
+    reset: () => ({
+      ...notebookState,
+    }),
+    resetError: (state) => ({
+      ...state,
+      error: null,
+    }),
+  },
+  extraReducers: (builder) => {
+    // START cluster
```
evans-g-crsj, line 29: comments like these // START cluster , in my opinion, bring very little value especially since your function methods are well-named.

src/store/notebook/slice.ts
```diff
@@ -0,0 +1,88 @@
...
+    builder.addCase(startNotebookCluster.pending, (state) => ({
+      ...state,
+      isLoading: true,
+      status: NotebookApiStatus.createInProgress,
+      error: undefined,
```
evans-g-crsj, line 34: (Optional) I think null would be better.

src/store/notebook/types.ts
```diff
@@ -0,0 +1,8 @@
+import { NotebookApiStatus } from 'services/api/notebook/model';
+
+export type initialState = {
+  url?: string | null;
```
evans-g-crsj, line 4: maybe that `null` is not necessary here since it s an option.

evans-g-crsj (changes requested):

src/views/Dashboard/components/DashboardCards/Notebook/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+import OpenInNewIcon from 'components/Icons/OpenInIcon';
+const { Text } = Typography;
+
+const Notebook = ({ id, className = '' }: DashboardCardProps) => {
+  const dispatch = useDispatch();
+  const { url, isLoading, error, status } = useNotebook();
+  const { groups } = useUser();
+  const { connectionStatus } = useFenceConnection();
+
+  const isConnectedToFences = hasOneFenceConnected(connectionStatus) || true;
```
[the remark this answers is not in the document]
lflangis: Just a test value that I forgot to reset. Should be without true

src/store/notebook/slice.ts
lflangis: Nah I like them, it divide all the calling block, easier to read when you have many reducers.

src/store/fenceConnection/slice.ts
```diff
@@ -22,7 +22,10 @@ export const FenceConnectionState: initialState = {
   loadingFences: [],
   fencesConnectError: [],
   fencesDisconnectError: [],
-  isFenceConnectionModalOpen: false,
+  modalConnectionParams: {
+    open: false,
+    onClose: undefined,
```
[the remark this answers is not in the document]
lflangis: putting null remove the concept of optional when declaring type. I don't think we have to fear it

```
export type TModalConnectionParams = {
  open: boolean;
  onClose?: () => void;
};
```

src/services/api/notebook/index.ts
lflangis: what would be the gain ?

src/components/uiKit/Fences/Modal/index.tsx
lflangis: the type are 'undefined' or a function, typescript will warms us if someone pass something else.

src/services/api/notebook/index.ts
evans-g-crsj: In this case, it is no big deal. But I think it is better to avoid raw strings in certain circumstances.

PR #3529, opened 2022-10-05
💄feat(profile): SKFP-471 add user profile view

lflangis:
# FEATURE

- closes [#471](https://d3b.atlassian.net/browse/SKFP-471)

## Description
The Profile view page that is currently in KF does not fit with the FerlabUI components. We will also simplify the fields that are asked to be filled out by the user. The Profile page layout will be the same as in INCLUDE but the fields will differ slightly. The Profile View page will display the information that was filled in by the user in the Profile Edit page. The specifications are indicated in the notion page below.

Add a new Header background image

Add alert banner if a user has their profile as private. Clicking on “settings” will redirect the user to the Profile Edit page
Acceptance Criterias

Profile Header
country_only
no_country_or_state
state_only

Community navigation (values are mockeD)
https://user-images.githubusercontent.com/65532894/194068069-e88343da-2646-407e-a789-3fee76e83b34.mp4

Live edit
https://user-images.githubusercontent.com/65532894/194067250-838b3bc6-b983-48da-84f1-4732ee31c701.mp4

Private profile alert
https://user-images.githubusercontent.com/65532894/194067352-90d9eaa0-6015-444c-9a8c-869b169a5670.mp4

## KNOW ISSUE

Diseases and studies interests containt duplicated values. Since there are all saved in the same fields `interests` in persona, it will create a duplicate in both fields. A ticket has been created to fix this issue later (we will change the interests to be all unique).

https://user-images.githubusercontent.com/65532894/194067075-659863d0-7066-4817-ac0f-208744a1f04e.mp4


src/components/Layout/Header/index.tsx
```diff
@@ -228,6 +228,7 @@ const Header = () => {
                     onClick: () => {
                       dispatch(personaActions.cleanLogout());
                       dispatch(userActions.cleanLogout());
+                      window.sessionStorage.clear();
```
evans-g-crsj, line 231: I think it is good but did you check with the product owner is that behaviour is ok (i think we ll lost the querybuilder stuff)?

src/store/persona/slice.ts
```diff
@@ -15,11 +21,22 @@ const personaSlice = createSlice({
   name: 'persona',
   initialState: PersonaState,
   reducers: {
-    cleanLogout: (_) => PersonaState,
+    cleanLogout: (state) => {
+      state.profile = undefined;
```
evans-g-crsj, line 25: I think that `null` is best suited. it is up to you.

src/views/Community/Member/index.tsx
```diff
@@ -79,28 +120,71 @@ const CommunityMember = () => {
                       }}
                     />
                   </Col>
-                  <Col span={24}>
-                    <Typography.Title level={5}>
-                      {intl.get('screen.memberProfile.usageTitle')}
-                    </Typography.Title>
-                  </Col>
+                  {(!!diseasesInterest?.length || !!studiesInterest?.length) && (
```
evans-g-crsj, line 123: I think it s best to use `> 0` vs `!!`

src/views/Community/components/MemberCard/index.tsx
```diff
@@ -15,26 +14,21 @@ interface OwnProps {
...
-        )}
-      </Space>
-    </Card>
-    // </Link>
+    <Link key={user._id} className={styles.memberLink} to={`/member/${user._id}`}>
+      <Card className={styles.memberCard}>
+        <Space direction="vertical" align="center">
+          <Gravatar className={styles.userGravatar} circle email={user.email || ''} />
+          <Typography.Title className={styles.memberCardName} level={5}>
+            <Highlighter textToHighlight={formatName(user) ?? ''} searchWords={[match]} />
```
evans-g-crsj, line 22: maybe `formatName` could handle the fallback value `''`

evans-g-crsj (approved):

src/components/Layout/Header/index.tsx
lflangis: query use the localstorage and not the sessionstorage, should have no impact :)

src/store/persona/slice.ts
lflangis: I think `null` should not be used and `undefined` should priorized https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

src/views/Community/components/MemberCard/index.tsx
lflangis: `  user.firstName && user.lastName ? `${user.firstName} ${user.lastName}` : user.email;` seems like it need to return the email by default, so maybe review it later when removing persona

PR #3541, opened 2022-10-12
🩹fix(dashboard): fix saved filters and set

lflangis:
# BUG

### DIFF
--- a/src/store/savedFilter/slice.ts
+++ b/src/store/savedFilter/slice.ts
@@ -1,6 +1,6 @@
 import { createSlice, PayloadAction } from '@reduxjs/toolkit';
-import { TUserSavedFilter } from 'services/api/savedFilter/models';
 import { initialState } from 'store/savedFilter/types';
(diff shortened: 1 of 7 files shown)


src/store/savedFilter/slice.ts
```diff
@@ -20,7 +20,17 @@ export const SavedFilterState: initialState = {
 };
 
 const sortByUpdateDate = (sets: TUserSavedFilter[]) =>
-  sets.sort((a, b) => (new Date(a.updated_date) < new Date(b.updated_date) ? 0 : -1));
+  sets.sort((a, b) => {
```
evans-g-crsj, line 23: 1. could you extract this sorting body for reuse
2. (to verify) I would be tempted to add a `.slice` before the sort: `sets.slice().sort(...)` to be sure we don t mutate `sets`. I don t believe it s an issue, but just to be sure...

evans-g-crsj (changes requested):

evans-g-crsj (approved):

PR #3559, opened 2022-10-19
:lispstick: Add translations in facets (variant search page)

evans-g-crsj:
image


src/graphql/utils/Filters.tsx
```diff
@@ -96,7 +96,7 @@ export const generateFilters = ({
   });
 
 const translateWhenNeeded = (group: string, key: string) =>
-  intl.get(`facets.options.${keyEnhance(key)}`).defaultMessage(keyEnhance(key));
+  intl.get(`facets.options.${group}.${keyEnhance(key)}`).defaultMessage(keyEnhance(key));
```
evans-g-crsj, line 99: the `group` param was not used in the function (see code before commit) which I think is an error. I had to add it in order to construct translation keys that are more complex and unique. For instance the property `D` could translate to `Damaging` or `Deleterious` so `group` is very much needed for without `group` we would have intl key `facets.options.D` but `D` can translate to many things. With `group` we now have `facets.options.groupA.D = Damaging` and or `facets.options.groupB.D = Deleterious`, for instance.

lflangis (approved):

PR #3576, opened 2022-11-08
:bug: Add correct tags to participants and variants queries (saved fi…

evans-g-crsj:
image

NOTE: this was used with a local version of `kf-api-arranger.` It may not work in QA but, at least, the ground is prepared.

### DIFF
--- a/src/services/api/savedFilter/models.ts
+++ b/src/services/api/savedFilter/models.ts
@@ -8,8 +8,8 @@ export type TUserSavedFilter = ISavedFilter & {
 };
 
 export enum SavedFilterTag {
-  PARTICIPANT = 'participant',
-  VARIANT = 'variants-exploration',
+  PARTICIPANTS_DExP = 'participants-data-exploration-page',
+  VARIANTS_VExP = 'variants-variants-exploration-page',
 }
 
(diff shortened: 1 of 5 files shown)


src/services/api/savedFilter/models.ts
```diff
@@ -8,8 +8,8 @@ export type TUserSavedFilter = ISavedFilter & {
 };
 
 export enum SavedFilterTag {
-  PARTICIPANT = 'participant',
-  VARIANT = 'variants-exploration',
+  PARTICIPANTS_DExP = 'participants-data-exploration-page',
```
lflangis, line 11: Is DexP volontary here ?

mobastille (approved):

src/services/api/savedFilter/models.ts
evans-g-crsj: Yes, for `Data Explore Page`. In case we need we need to support participants filter elsewhere in the app.

lflangis: Should our enum be un `PascalCase`? https://www.typescriptlang.org/docs/handbook/enums.html

PR #3588, opened 2022-11-16
🐛fix(variant): SKFP-377 add share icon

lflangis:
# BUG

- closes #[377](https://d3b.atlassian.net/browse/SKFP-377)

## Description
- Add missing shared icon
- Fix Dash dashboard card to redirect to the correct page
- Remove FilterWithActions to use core component

## Screenshot

Before
image

After
image


src/views/Dashboard/components/DashboardCards/SavedFilters/ListItem/index.tsx
```diff
@@ -39,22 +40,17 @@ const SavedFiltersListItem = ({ data }: OwnProps) => {
...
-            </Text>
-          ),
-          onClick: () =>
-            setQueryBuilderState(FILTER_TAG_QB_ID_MAPPING[data.tag], {
-              active: data.queries[0].id,
-              state: data.queries,
-            }),
+        onClick={() => {
+          history.push(
+            `${FILTER_TAG_PAGE_MAPPING[data.tag]}?${FILTER_ID_QUERY_PARAM_KEY}=${data.id}`,
```
evans-g-crsj, line 45: should we use an api such as searchParams to build url params? That would take care of correctly escaping and such.

src/views/Dashboard/components/DashboardCards/SavedFilters/ListItem/index.tsx
```diff
@@ -39,22 +40,17 @@ const SavedFiltersListItem = ({ data }: OwnProps) => {
...
+          );
+
+          setQueryBuilderState(FILTER_TAG_QB_ID_MAPPING[data.tag], {
+            active: data.queries[0].id,
```
evans-g-crsj, line 49: do we expect `data.queries[0]` to always exist? If yes, leave it as it is if not, add an `?`

evans-g-crsj (approved):

src/views/Dashboard/components/DashboardCards/SavedFilters/ListItem/index.tsx
lflangis: I sync with Includes for that one. Maybe reworking those when needed since it could impact all our project

PR #3599, opened 2022-11-23
🐛fix(studies): SKFP-315 Show results when clicking the Studies graph

mobastille:
- closes [515](https://d3b.atlassian.net/browse/SKFP-515)

## Description

- Fixed the Studies graph that didn't show any results when clicked because the wrong key was sent.
- Cleaned up some variables that were not used.
- Created an enum to handle summary fields

## Screenshot
Before:
515_before

After:
515_after

### DIFF
--- a/src/views/DataExploration/components/PageContent/tabs/Summary/DemographicGraphCard/index.tsx
+++ b/src/views/DataExploration/components/PageContent/tabs/Summary/DemographicGraphCard/index.tsx
@@ -1,24 +1,24 @@
-import { Col, Row } from 'antd';
-import PieChart from 'components/uiKit/charts/Pie';
-import { toChartData } from 'utils/charts';
 import intl from 'react-intl-universal';
-import GridCard from '@ferlab/ui/core/view/v2/GridCard';
-import { INDEXES } from 'graphql/constants';
-import { useHistory } from 'react-router-dom';
-import { ArrangerValues } from '@ferlab/ui/core/data/arranger/formatting';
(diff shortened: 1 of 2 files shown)


src/views/DataExploration/components/PageContent/tabs/Summary/DemographicGraphCard/index.tsx
```diff
@@ -93,7 +92,7 @@ const DemographicsGraphCard = ({ id, className = '' }: OwnProps) => {
               <PieChart
                 title={intl.get('screen.dataExploration.tabs.summary.demographic.sexTitle')}
                 data={sexData}
-                onClick={(datum) => addToQuery('sex', datum.id as string, history)}
+                onClick={(datum) => addToQuery('sex', datum.id as string)}
```
evans-g-crsj, line 95: do the fields `sex`, `race` and `ethnicity` exist in an enum? If not, do you think we should create one?

evans-g-crsj (approved):

lflangis (approved):

src/views/DataExploration/components/PageContent/tabs/Summary/DemographicGraphCard/index.tsx
mobastille: Yes I think we should definitely avoid passing a hardcoded string as a key. Not sure if that follows what you had in mind, but I created a SUMMARY_KEY enum to handle the possible keys we use in the Summary section.

PR #3625, opened 2022-12-12
🐛fix(data-exploration): SKFP-560 Study column sorting

mobastille:
# BUG

- closes [SKFP-562](https://d3b.atlassian.net/browse/SKFP-560)

## Description

Refact of the ```formatQuerySortList()``` function. Made sure it still works everywhere it's used.

Only thing to keep in mind is that when creating a table, the 'key' property will now always be used as the ```field``` value sent to the backend for sorting (i.e. we can use ```key: 'study.study_code'``` for nested data) and the ```dataIndex```property will only be used to render the cell content.

Before, `dataIndex` would become ```sorter.field``` in the ```formatQuerySortList()``` function, which created inconsistent ways of using ```key``` vs ```dataIndex``` and would sometimes result in breaking the sorting because the wrong key was sent to the backend. 

## Screenshot
Before:
560_before

After:
560_after

### DIFF
--- a/src/utils/helper.test.tsx
+++ b/src/utils/helper.test.tsx
@@ -3,7 +3,7 @@ import { SorterResult } from 'antd/lib/table/interface';
 
 import { formatQuerySortList, getOrderFromAntdValue } from './helper';
 
-describe('getOrderFromAntdValue()', () => {
+describe(`${getOrderFromAntdValue.name}()`, () => {
   test('should return SortDirection.Asc for "ascend"', () => {
     expect(getOrderFromAntdValue('ascend')).toBe(SortDirection.Asc);
   });
@@ -13,7 +13,7 @@ describe('getOrderFromAntdValue()', () => {
   });
 });
 
-describe('formatQuerySortList()', () => {
+describe(`${formatQuerySortList.name}()`, () => {
   it('should handle a single sorter', () => {
     const sorter1: SorterResult<any> = {
       columnKey: 'field1',
--- a/src/utils/helper.test.tsx
+++ b/src/utils/helper.test.tsx
@@ -1,7 +1,7 @@
 import { SortDirection } from '@ferlab/ui/core/graphql/constants';
 import { SorterResult } from 'antd/lib/table/interface';
 
-import { convertToArray, formatQuerySortList, getOrderFromAntdValue } from './helper';
+import { formatQuerySortList, getOrderFromAntdValue } from './helper';
 
 describe('getOrderFromAntdValue()', () => {
   test('should return SortDirection.Asc for "ascend"', () => {
@@ -10,18 +10,6 @@ describe('getOrderFromAntdValue()', () => {
 
   test('should return SortDirection.Desc for any other value', () => {
     expect(getOrderFromAntdValue('foo')).toBe(SortDirection.Desc);
-  });
-});
-
-describe('convertToArray()', () => {
-  it('should return an array if input is an array', () => {
-    const array = [1, 2, 3];
-    expect(convertToArray(array)).toEqual(array);
-  });
-
-  it('should return an array if input is not an array', () => {
-    const value = 'Hello World';
-    expect(convertToArray(value)).toEqual([value]);
   });
 });
 
--- a/src/utils/helper.test.tsx
+++ b/src/utils/helper.test.tsx
@@ -1,0 +1,51 @@
+import { SortDirection } from '@ferlab/ui/core/graphql/constants';
+import { SorterResult } from 'antd/lib/table/interface';
+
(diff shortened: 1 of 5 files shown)


src/utils/helper.ts
```diff
@@ -25,7 +23,9 @@ export const formatQuerySortList = (sorter: SorterResult<any> | SorterResult<any
   );
 
   const r = sorters.map((sorter) => ({
-    field: (sorter.field?.toString()! || sorter.columnKey?.toString()!).replaceAll('__', '.'),
+    field: (
+      sorter.field?.toString()!.replaceAll(',', '.') || sorter.columnKey?.toString()!
```
evans-g-crsj, line 27: why using `!` while using `or` operand and `?` ?

src/utils/helper.ts
```diff
@@ -11,9 +11,7 @@ export const scrollToTop = (scrollContentId: string) =>
 
 export const orderCardIfNeeded = (cards: TSortableItems[], userCardConfig: string[] | undefined) =>
   userCardConfig
-    ? cards.sort((a, b) => {
-        return userCardConfig.indexOf(a.id) > userCardConfig.indexOf(b.id) ? 1 : -1;
-      })
+    ? cards.sort((a, b) => (userCardConfig.indexOf(a.id) > userCardConfig.indexOf(b.id) ? 1 : -1))
```
evans-g-crsj, line 14: Beware that `sort` modifies the original array. Are you sure that `cards` are not modify in the parent component?

evans-g-crsj (changes requested):

src/utils/helper.ts
mobastille: I did not change anything here, it's just the linter that reorganized the code

src/utils/helper.ts
mobastille: Hmm I'm not sure I only added the ```.replaceAll(',', '.')``` part to handle array dataIndexes: ```dataIndex: ['study', 'study_code'],``` gives: ```'study,study_code'``` so I change it to ```study.study_code``` so we send the correct field key to the backend for sorting.

But I agree it makes no sense to have ```?```, ```!```and ```||```all at the same time, I just didn't want to modify the existing logic. But I could take a look across the different tables with sorting and see what's going on

src/utils/helper.ts
mobastille: Added ```slice()``` as discussed

src/utils/helper.ts
```diff
@@ -11,27 +11,22 @@ export const scrollToTop = (scrollContentId: string) =>
...
     : cards;
 
 export const getOrderFromAntdValue = (order: string): SortDirection =>
   order === 'ascend' ? SortDirection.Asc : SortDirection.Desc;
 
-export const formatQuerySortList = (sorter: SorterResult<any> | SorterResult<any>[]) => {
-  const sorters = (isArray(sorter) ? sorter : [sorter]).filter(
-    (sorter) => !!sorter.column || !!sorter.order,
-  );
+export const convertToArray = <T>(value: T | T[]) => (isArray(value) ? value : [value]);
```
evans-g-crsj, line 22: Personally, I think `[l].flat()` is more idiomatic. Besides, I don't see the point of making a function out of it.

evans-g-crsj (changes requested):

src/utils/helper.test.tsx
```diff
@@ -0,0 +1,51 @@
+import { SortDirection } from '@ferlab/ui/core/graphql/constants';
+import { SorterResult } from 'antd/lib/table/interface';
+
+import { convertToArray, formatQuerySortList, getOrderFromAntdValue } from './helper';
+
+describe('getOrderFromAntdValue()', () => {
```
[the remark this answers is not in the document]
mobastille: good to know, thanks!

src/utils/helper.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  test('should return SortDirection.Desc for any other value', () => {
+    expect(getOrderFromAntdValue('foo')).toBe(SortDirection.Desc);
+  });
+});
+
+describe('convertToArray()', () => {
+  it('should return an array if input is an array', () => {
+    const array = [1, 2, 3];
+    expect(convertToArray(array)).toEqual(array);
```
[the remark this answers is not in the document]
mobastille: I removed this function and its tests after all

src/utils/helper.ts
mobastille: Removed the convertToArray function and used ```[l].flat()``` instead

evans-g-crsj (approved):

PR #3626, opened 2022-12-13
:bugs: Fix wording for files type in set dropdown (filess vs files)

evans-g-crsj:
image

### DIFF
--- a/src/views/DataExploration/components/SetsManagementDropdown/index.tsx
+++ b/src/views/DataExploration/components/SetsManagementDropdown/index.tsx
@@ -88,6 +88,9 @@ const itemIcon = (type: string) => {
   }
 };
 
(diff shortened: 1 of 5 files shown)


src/views/DataExploration/components/SetsManagementDropdown/index.tsx
```diff
@@ -88,6 +88,9 @@ const itemIcon = (type: string) => {
   }
 };
 
+const singularizeFilesSetType = (type: string) =>
+  type === SetType.FILES ? type.slice(0, -1) : type;
+
```
evans-g-crsj, line 93: Hacky, but we rely on `SetType` enum for the wording and did not know how to handle this issue differently without changing a lot of code.

mobastille (approved):

PR #3634, opened 2022-12-16
🐛fix(saved-sets): SKFP-568 Fix QueryPill title and file set query

mobastille:
# BUG 

- closes #[SKFP-568](https://d3b.atlassian.net/browse/SKFP-568)

## Description

- Fix the QueryPill title for saved sets
- Fix the file saved set query

## Screenshot 
Before:
568_before

After:
568_after

### DIFF
--- a/src/store/savedSet/index.ts
+++ b/src/store/savedSet/index.ts
@@ -1,13 +1,18 @@
 import { useSelector } from 'react-redux';
-import { savedSetSelector } from './selector';
-import EnvironmentVariables from '../../helpers/EnvVariables';
+import { singularizeFilesSetType } from 'views/DataExploration/components/SetsManagementDropdown';
+
 import { SetType } from 'services/api/savedSet/models';
+
+import EnvironmentVariables from '../../helpers/EnvVariables';
+
+import { savedSetSelector } from './selector';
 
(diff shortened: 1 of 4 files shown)


src/store/savedSet/index.ts
```diff
@@ -1,13 +1,18 @@
...
+
+import { savedSetSelector } from './selector';
+
 export type { initialState as SavedSetInitialState } from './types';
 export { default, SavedSetState } from './slice';
 
 export const MAX_LENGTH_NAME = 50;
-export const getSetFieldId = (type: SetType) => `${type}_facet_ids.${type}_fhir_id_1`;
+export const getSetFieldId = (type: SetType) =>
+  `${singularizeFilesSetType(type)}_facet_ids.${singularizeFilesSetType(type)}_fhir_id_1`;
```
mobastille, line 15: I don't really like this solution but I don't see any other way, also was wondering if ```singularizeFilesSetType```should be moved to a utils file somewhere else

evans-g-crsj: You are right, but I think it is fine. Perhaps we could remove the dependence of `type` on the way we name our indices.

evans-g-crsj (approved):

PR #3655, opened 2023-01-13
✨feat(querybar): SKFP-596 add an remoteComponent properties

lflangis:
#FEATURE

- closes #[596](https://d3b.atlassian.net/browse/SKFP-596)

## Description
 A user opening the Query Pill dropdown for an HPO or MONDO term should automatically open the corresponding tree instead of the dropdown list. 

## Screenshot
Before
image

After
Peek 2023-01-13 14-00


src/store/remote/selector.ts
```diff
@@ -0,0 +1,8 @@
+import { RootState } from 'store/types';
+import { initialState } from 'store/remote/types';
+
+export type TRemoteProps = initialState;
+
+export const remoteSelector = (state: RootState) => {
```
evans-g-crsj, line 6: `const remoteSelector = (state: RootState) => state.remote`

src/views/DataExploration/components/TreeFacet/TreeFacetModal.tsx
```diff
@@ -0,0 +1,361 @@
+import { ReactElement, useEffect, useRef, useState } from 'react';
```
evans-g-crsj, line 1: I think it is legacy code. If not, I would try to:

- breakdown the file (relatively large)
- be a bit more functional (that s a personal taste, but a paradigm we use quite a bit)

lflangis: It's legacy, I prefer to not touch it

PR #3678, opened 2023-02-15
:building_construction: Allow to make a production build

evans-g-crsj:
Currently, we cannot use NODE_ENV='production'. This commit fixes the issues. However, some testing libraries had to be put in the dependencies. That should not be the case.


package.json
```diff
@@ -63,6 +63,19 @@
...
+    "@types/react-router-dom": "^5.3.2",
+    "@types/d3": "^7.1.0",
+    "@types/lodash": "^4.14.178",
+    "@types/node": "^16.11.7",
+    "@types/react": "^17.0.34",
+    "@types/react-dom": "^17.0.11",
+    "@types/react-highlight-words": "^0.16.4",
+    "@types/react-router": "^5.1.17",
+    "@types/redux-logger": "^3.0.9",
+    "@testing-library/react": "^12.1.2",
```
evans-g-crsj, line 77: Unfortunately, I had no choice to this. Suggestion: two ts config files: 1 for prod, 1 for dev and excluding test files in production.

mobastille (approved):

PR #3706, opened 2023-03-22
:chart_with_upwards_trend: No index, no follow for beta site

evans-g-crsj:
We don't want beta portal to be listed by search engine at this moment.

### DIFF
--- a/public/_headers
+++ b/public/_headers
@@ -1,2 +1,3 @@
 /*
   X-Frame-Options: SAMEORIGIN
+  X-Robots-Tag: noindex, nofollow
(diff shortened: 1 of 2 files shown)


public/_headers
evans-g-crsj, line 1: ref: https://yoast.com/prevent-site-being-indexed/

public/index.html
evans-g-crsj, line 1: ref: https://yoast.com/prevent-site-being-indexed/

lflangis (approved):

alubneuski (approved):

PR #3790, opened 2023-09-28
fix: skfp-719 fix auth studies sqon 3

francisl:
# BUG 

closes [#3242](https://d3b.atlassian.net/browse/SKFP-719)

Description
This sqon is not well contructed. It missing a content array wrapping it. See PR that was done for include to fix the same issue:
## Validation

- [ ] Code Approved
- [ ] Test Coverage
- [ ] QA Done

### DIFF
--- a/src/store/fenceStudies/index.ts
+++ b/src/store/fenceStudies/index.ts
@@ -1,7 +1,9 @@
-import { FENCE_CONNECTION_STATUSES, FENCE_NAMES } from 'common/fenceTypes';
-import { isEmpty } from 'lodash';
 import { useSelector } from 'react-redux';
+import isEmpty from 'lodash/isEmpty';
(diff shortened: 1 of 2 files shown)


AltefrohneGaelle (approved):

src/store/fenceStudies/index.ts
```diff
@@ -1,7 +1,9 @@
-import { FENCE_CONNECTION_STATUSES, FENCE_NAMES } from 'common/fenceTypes';
-import { isEmpty } from 'lodash';
 import { useSelector } from 'react-redux';
+import { isEmpty } from 'lodash';
```
ethienneroy, line 2: peut-etre mieux d'importer seulement isEmpty plutot que toute la lib

PR #3830, opened 2023-10-25
:typo: Rename fields due to data change for specimens

evans-g-crsj:


src/graphql/participants/queries.ts
```diff
@@ -168,8 +168,8 @@ export const GET_PARTICIPANT_ENTITY = gql`
                             dbgap_consent_code
                             fhir_id
                             laboratory_procedure
-                            ncit_anatomy_site_id
-                            ncit_id_tissue_type
+                            collection_ncit_anatomy_site_id
+                            # TODO ncit_id_tissue_type
```
evans-g-crsj, line 172: we don t know what to do with it yet

PR #3866, opened 2023-12-11
Publish/beta prd 1.5.0

francisl:
Update Beta-Prd


evans-g-crsj (changes requested):
The branch should has 2.0 has origin. To be validated with Luc-F and Gaelle

PR #3994, opened 2024-05-16
fix: #SKFP-000 fix google analytics for legacy

celinepelletier:
https://www.npmjs.com/package/react-ga4

```
// Simply replace `react-ga` with `react-ga4` and remove `ReactGA.pageview()`
// import ReactGA from "react-ga";
```

### DIFF
--- a/src/services/analyticsTracking/googleAnalytics.js
+++ b/src/services/analyticsTracking/googleAnalytics.js
@@ -1,4 +1,4 @@
-import ReactGA from 'react-ga';
+import ReactGA from 'react-ga4';
 import isObject from 'lodash/isObject';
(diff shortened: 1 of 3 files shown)


src/services/analyticsTracking/googleAnalytics.js
```diff
@@ -199,7 +199,7 @@ export const trackPageView = (page, options = {}) => {
     page,
     ...options,
   });
-  ReactGA.pageview(page);
+  // ReactGA.pageview(page);
```
aperron-ferlab, line 202: why not remove? Maybe add comment

aperron-ferlab (approved):

